PR #104, opened 2020-10-27
Fix gRPC client hanging issue in error scenarios

daneshk:
## Purpose
Resolves: ballerina-platform/ballerina-standard-library#327


chamil321:
@daneshk Shouldn't we change L164 and L189 of the following file also? That is the location where the HTTP/2 client reads inbound response headers.

https://github.com/ballerina-platform/module-ballerina-http/blob/master/http-native/src/main/java/org/ballerinalang/net/transport/contractimpl/sender/states/http2/ReceivingHeaders.java

chamil321 (approved):

PR #132, opened 2020-12-09
Add query param binding support

chamil321:
```ballerina
type BasicType boolean|int|float|decimal|string;
public type QueryParamType ()|BasicType|BasicType[];
```
> The query params can have multiple values. In the presence of multiple such values,  If the user has specified the param as an array type, then all values will return. If not the first param values will be returned.

> If the query parameter is not defined in the function signature, then the query param binding does not happen. 
> If a query param of the request URL has no corresponding parameter in the resource function, then that param is ignored. 
> When the parameter is defined in the function, but if there is no such query param in the URL, that request will lead to a 400 error response unless the type is nilable (string|()). 
> If the query param type casting fails, it will result in a 500 error response.

Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/593

## Sample
```ballerina
import ballerina/http;

service /query on new http:Listener(9090) {
    resource function get q1(int a, string[] b, http:Caller caller, float[]? c, boolean? d, decimal e) {
        json responseJson = { iValue: a, sValue: b, fValue: c, bValue: d, dValue: e };
        checkpanic caller->respond(<@untainted json> responseJson);
    }
}
```

## Automation tests
 - Unit tests 
   > Added


http-native/src/main/java/org/ballerinalang/net/http/signature/ParamHandler.java
```diff
@@ -0,0 +1,233 @@
...
+            }
+        }
+        if (queryParams.isNotEmpty()) {
+            getOtherParamList().add(this.queryParams);
+        }
+
+
+
+
+//        if (resource.getEntityBodyAttributeValue() == null ||
```
chamil321, line 122: Commented out temporarily until the annotation support

Bhashinee (approved):

PR #163, opened 2021-01-15
Add support to retry secure client when OAuth2 client failed

ldclakmal:
## Purpose
This PR add capability to retry the HTTP client automatically if it receives an 401 response while it is secured with OAuth2. Further, the response code also can be configured.

Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/17

Related to https://github.com/ballerina-platform/module-ballerina-oauth2/pull/52
Related to https://github.com/ballerina-platform/ballerina-standard-library/issues/584


ldclakmal:
This scenario is not valid since the token refresh in handled internally when the cached token is expired. Therefore this situation cannot be happened.

PR #211, opened 2021-02-10
 Introduce error inheritance

chamil321:


anupama-pathirage:
@chamil321 Is this PR still valid? Seems bit old to keep open. Please resolve conflicts and get this merged asap.

chamil321:
> @chamil321 Is this PR still valid? Seems bit old to keep open. Please resolve conflicts and get this merged asap.

Sure, I Will get it merged soon. This was on hold a bit until a lang fix(ballerina-platform/ballerina-lang#28563)

http-ballerina/http_errors.bal
```diff
@@ -14,152 +14,156 @@
...
+    error[] httpActionErr?;
+};
+
+// Level 1
+# Defines the common error type for the module
+public type Error distinct error;
+
+// Level 2
+# Defines the possible listener error types
+public type ListenerError distinct Error;
```
ThisaruGuruge, line 37: Do we need separate `ListenerError` and `ClientError`?

ayeshLK (approved):

http-ballerina/http_errors.bal
chamil321: Thought it would create a clear separation between client and listener errors. 
https://docs.google.com/spreadsheets/d/1bkOn7TahMyz22HC1XlE8x7-ByIZV1AZf6eDC4sRXEXM/edit?usp=sharing

ThisaruGuruge: In a recent discussion, we agreed to not to have a separation between client and listener errors, since they cannot occur at the same time. (Not sure if that applies to the HTTP scenario though).

shafreenAnfar: Yeah, I was thinking the same. IMO, no point in having both even in HTTP.

chamil321: From a single method POV, there is no such occurrence where both client and listener errors are returned. But there could be situations such as `trap` and `do {} on fail{}` where you capture both types. In such cases, this separation help user to isolate the root cause up to a certain extent. 

I have introduces common error `http:Error` also, so the user has the option to use it when the separation is not needed.

shafreenAnfar (approved):

PR #231, opened 2021-02-17
Marking `cookie.toStringValue()` function public, so the users can directly set cookie object as a header without calling addCookie()

Bhashinee:
## Purpose
Marked getting the string value of a cookie function public, so that this can be used in WebSocket-like connections to set the cookie object as a header as we don't expose the http:Response of the initial request to the users to call `addCookie()` function.

### DIFF
--- a/http-ballerina/cookie.bal
+++ b/http-ballerina/cookie.bal
@@ -121,8 +121,10 @@ public class Cookie {
         return true;
     }
 
-    // Gets the Cookie object in its string representation to be used in the ‘Set-Cookie’ header of the response.
-    function toStringValue() returns string {
+    # Gets the Cookie object in its string representation to be used in the ‘Set-Cookie’ header of the response.
+    #
+    # + return  - The string value of the ‘http:Cookie’
+    public function toStringValue() returns string {
         string setCookieHeaderValue = "";
         var temp1 = self.name;
         var temp2 = self.value;


chamil321:
What is the limitation here? you don't have the http:Response object?

Bhashinee:
> What is the limitation here? you don't have the http:Response object?
This is not a limitation in HTTP. 
In the WebSocket upgrade service we are not allowing the users to create or set anything to the `http:Response`, it is created internally as it has to be created setting relevant headers and content. But users can specify custom headers to be set to that internally created response. Users are only allowed to pass a custom headers map.

So if the user wants to create a Cookie and send it as header, if we make this function public, user can do something like this,

```http:Cookie cookie = new ("username", "name");
map<string> customHeaders = {"Set-Cookie" : cookie.toStringValue()};```

chamil321:
> > What is the limitation here? you don't have the http:Response object?
> 
> This is not a limitation in HTTP.
> In the WebSocket upgrade service we are not allowing the users to create or set anything to the `http:Response` which is the upgrade response, it is created internally as it has to be created setting specific headers. But users can specify custom headers to be set to that internally created response. Users are only allowed to pass a custom headers map.
> 
> So if the user wants to create a Cookie and send it in a header, if we make this function public, user can do something like this,
> 
> ```
> http:Cookie cookie = new ("username", "name");
> cookie.domain = "xyz";
> cookie.path = "bee";
> map<string> customHeaders = {"Set-Cookie" : cookie.toStringValue()};
> ```

+1. Cookie object should have a function like this to be used independently.

chamil321 (approved):

PR #244, opened 2021-02-23
Fix anydata equality check failure

chamil321:


http-ballerina/http_request.bal
```diff
@@ -195,7 +195,13 @@ public class Request {
     #
     # + return - Returns true if the client expects a `100-continue` response
     public isolated function expects100Continue() returns boolean {
-        return <@untainted> (self.hasHeader(EXPECT) ? self.getHeader(EXPECT) == "100-continue" : false);
+        if (self.hasHeader(EXPECT)) {
+            string|error value = self.getHeader(EXPECT);
+            if (value is string && value == "100-continue") {
+                return true;
+            }
```
ldclakmal, line 202: Can this be simplified as follows?
```suggestion
            return value is string && value == "100-continue";
```

ldclakmal (approved):

PR #230, opened 2021-02-17
Codecov Report Publishing

trajinthan:
## Purpose

- Introducing `Codecov` code coverage report publishing to the `http` repository

## Goals

- Generating a code coverage report that is published via `Codecov` and triggered via GitHub actions of build and pull request.
- `Codecov` bot commenting on PRs

## Approach

- Custom execution file, java and ballerina class files, Sourcecode is provided to `jacoco` to generate an XML code coverage report.
- A new job is inserted to `Build` and `Pull request` GitHub actions that publishes the XML report to `Codecov`.
- These jobs trigger on merge to master or pull request and CodeCov generates a descriptive code coverage report.
- `Codecov` also adds comments to any PR made based on the code coverage changes between commits

PR #246, opened 2021-02-24
fix codecoverage workflow

trajinthan:
## Purpose
Refer #230

### DIFF
--- a/.github/workflows/build-master.yml
+++ b/.github/workflows/build-master.yml
@@ -26,7 +26,8 @@ jobs:
           packageUser: ${{ secrets.BALLERINA_BOT_USERNAME }}
           packagePAT: ${{ secrets.BALLERINA_BOT_TOKEN }}
         run: |
-          ./gradlew publish
+            ./gradlew publish
+            ./gradlew codeCoverageReport
       - name: Archive Error Log
         uses: actions/upload-artifact@v2
         if: failure()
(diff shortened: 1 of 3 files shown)


codecov.yml
```diff
@@ -1,2 +1,5 @@
 fixes:
   - "ballerina/http/*/::../http-ballerina/"
+
+ignore:
+  - "http-ballerina-tests"
```
ThisaruGuruge, line 5:
```suggestion
  - "http-ballerina-tests"
  
```

.github/workflows/build-master.yml
```diff
@@ -26,7 +26,8 @@ jobs:
           packageUser: ${{ secrets.BALLERINA_BOT_USERNAME }}
           packagePAT: ${{ secrets.BALLERINA_BOT_TOKEN }}
         run: |
-          ./gradlew publish
+            ./gradlew publish
+            ./gradlew codeCoverageReport
```
ThisaruGuruge, line 30: Are we going to use two steps? I see this is different from [what GraphQL does.](https://github.com/ballerina-platform/module-ballerina-graphql/pull/98) Is there a reason for that?

trajinthan: yes. There is a difference. GraphQL uses the xml generated by testerina so we don't need a seperate task there. In http we can't use that xml because that xml only contains the `http-ballerina-tests` module. so we need a task to generate xml which includes `http-ballerina` module.

ThisaruGuruge (approved):
LGTM

PR #249, opened 2021-02-25
update codecov task dependency

trajinthan:
#246

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -247,7 +247,6 @@ task build {
 }
 
 task codeCoverageReport(type: JacocoReport) {
-    dependsOn('build')
     dependsOn('http-ballerina-tests:codeCoverageReportBal')
     dependsOn('http-native:codeCoverageReportJava')
 }
--- a/build.gradle
+++ b/build.gradle
@@ -247,6 +247,7 @@ task build {
 }
 
 task codeCoverageReport(type: JacocoReport) {
+    dependsOn('build')
     dependsOn('http-ballerina-tests:codeCoverageReportBal')
     dependsOn('http-native:codeCoverageReportJava')
 }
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -247,6 +247,7 @@ task build {
 }
 
 task codeCoverageReport(type: JacocoReport) {
+    dependsOn('build')
```
ThisaruGuruge, line 250: Shouldn't we update the other task dependencies as well?

i.e: `'http-ballerina-tests:codeCoverageReportBal'` should depends on the `http-ballerina:build` right?

trajinthan: yes, that's my fault, need to update `http-ballerina-tests:codeCoverageReportBal`dependency. I think i can remove the build dependency from here. am I right?

ThisaruGuruge: Yes, if we add dependencies for the other two tasks, we can remove the dependency here.

http-ballerina-tests/build.gradle
```diff
@@ -239,6 +239,7 @@ task extractBallerinaClassFiles(type: Copy) {
 }
 
 task codeCoverageReportBal(type: JacocoReport) {
+    dependsOn('http-ballerina:build')
```
ThisaruGuruge, line 242:
```suggestion
    dependsOn(':http-ballerina:build')
```

http-native/build.gradle
```diff
@@ -154,6 +154,7 @@ task extractJavaClassFiles(type: Copy) {
 }
 
 task codeCoverageReportJava(type: JacocoReport) {
+    dependsOn(':http-ballerina:build')
```
ThisaruGuruge, line 157: This should depend on `:http-native:build`.

trajinthan: yes, sorry i am in a hurry.

ThisaruGuruge (approved):

PR #321, opened 2021-04-19
Close obs context manually after adding status code

tharinduwijewardane:
## Purpose
Manually closing the observer context after adding status code to resource span. It is required when the user has put a return statement instead of a caller->respond method in the resource function.

Lang PR: https://github.com/ballerina-platform/ballerina-lang/pull/30074

## Checklist
- [x] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests -- will be added with https://github.com/wso2-enterprise/choreo/issues/3437


http-native/src/main/java/org/ballerinalang/net/http/HttpCallableUnitCallback.java
```diff
@@ -47,6 +51,11 @@
     public void notifySuccess(Object result) {
         if (result == null) { // handles nil return and end of resource exec
             requestMessage.waitAndReleaseAllEntities();
+            ObserverContext observerContext
```
chamil321, line 54: Shouldn't we check whether `ObserveUtils.isObservabilityEnabled()` before looking for the `ObserverContext`?

tharinduwijewardane: `ObserverContext` will be null when observability is disabled hence that check is redundant. Should I add it anyway?

chamil321: Yeah, we can get rid of the map lookup and the cast if check it right? Shall we do that by considering the potential perf gain?

tharinduwijewardane: Fixed

chamil321 (approved):

http-native/src/main/java/org/ballerinalang/net/http/HttpCallableUnitCallback.java
```diff
@@ -47,6 +51,13 @@
     public void notifySuccess(Object result) {
         if (result == null) { // handles nil return and end of resource exec
             requestMessage.waitAndReleaseAllEntities();
+            if (ObserveUtils.isObservabilityEnabled()) {
+                ObserverContext observerContext
+                        = (ObserverContext) requestMessage.getProperty(OBSERVABILITY_CONTEXT_PROPERTY);
+                if (observerContext != null) {
+                    ObserveUtils.stopObservationWithContext(observerContext);
```
nadundesilva, line 58: Don't we need to check whether the observation should be manually closed (`if (observerContext.isManuallyClosed())`) here before closing it ? (Same in the `sendFailureResponse` method as well)

tharinduwijewardane: `observerContext` being present in the request message means it should be manually closed. Should I explicitly check it?

nadundesilva: Ahh yes. It should be enough :+1:

nadundesilva (approved):

PR #443, opened 2021-06-09
Return decoded form data from the getFormParam function

chamil321:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/1194

## Checklist
- [x] Linked to an issue
- [ ] Updated the changelog
- [x] Added tests


http-ballerina/http_request.bal
```diff
@@ -389,6 +389,8 @@ public class Request {
                             string value = entries[entryIndex].substring(index + 1, size);
                             value = value.trim();
                             if (value != "") {
+                                name = check decode(name, CHARSET_UTF_8);
+                                value = check decode(value, CHARSET_UTF_8);
```
ayeshLK: * Do we have to do url-decode for each entry? Can't we just decode the whole payload once and then split it with `&` ?

http-ballerina/http_request.bal
```diff
@@ -741,3 +743,7 @@ isolated function externRequestHasHeader(Request request, string headerName, Hea
     'class: "org.ballerinalang.net.http.nativeimpl.ExternHeaders",
     name: "hasHeader"
 } external;
+
+isolated function decode(string str, string charset) returns string|GenericClientError = @java:Method {
+    'class: "org.ballerinalang.net.uri.nativeimpl.Decode"
+} external;
```
ayeshLK: * Why can't we use `ballerina-url` package for encode/decode ?

http-ballerina/http_request.bal
```diff
@@ -389,7 +390,7 @@ public class Request {
                             string value = entries[entryIndex].substring(index + 1, size);
                             value = value.trim();
                             if (value != "") {
-                                parameters[name] = value;
+                                parameters[check decode(name)] = check decode(value);
```
ayeshLK, line 393: * Do we have to do url-decode for each entry? Can't we just decode the whole payload once and then split it with `&` ?

ldclakmal (approved):

ayeshLK (approved):

PR #461, opened 2021-06-15
Fix dispatching logic to handle repeated and same positioned path templates

chamil321:
## Examples
```ballerina
    // Case 1: `id` is present in 1 & 2 resources and positioned as a simpleString expression
    resource function get path/[string id]() returns string {
        return "a-" + id;
    }

    resource function get path/[string note]/aa/[string id]() returns string {
        return "b-" + note + " " + id;
    }

    // Case 2: `aaa` and `bbb` are present in both resource and in twisted positions
    resource function get [string aaa]/go/[string bbb]() returns string {
        return "c-" + aaa + " " + bbb;
    }

    resource function get [string bbb]/go/[string aaa]/[string ccc]() returns string {
        return "d-" + aaa + " " + bbb + " " + ccc;
    }

    // Case 3: rest param present along with path param
    resource function get bar/[string bbb]/go/[string... a]() returns string {
        return "e-" + a[0] + " " + bbb;
    }

    // Case 4: rest param has same var as 2nd resource
    resource function get path/[string id]/aa/[string... note]() returns string {
        return "f-" + note[1] + " " + id;
    }
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,12 @@ This file contains all the notable changes done to the Ballerina HTTP package th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to 
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+###Fixed
+- [Fix the limitation of not supporting different API resources with same end URL Template](https://github.com/ballerina-platform/ballerina-standard-library/issues/1095)
+- [Fix dispatching failure when same path param identifiers exist in a different order](https://github.com/ballerina-platform/ballerina-standard-library/issues/342)
+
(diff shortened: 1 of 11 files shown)


changelog.md
```diff
@@ -4,6 +4,12 @@ This file contains all the notable changes done to the Ballerina HTTP package th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to 
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+###Fixed
```
ldclakmal, line 9:
```suggestion
### Fixed
```

ldclakmal (approved):

PR #474, opened 2021-06-18
Add trailer header check before processing headers

TharmiganK:
## Purpose
Fixes [`ballerina-platform/ballerina-standard-library#808`](https://github.com/ballerina-platform/ballerina-standard-library/issues/808)

### DIFF
--- a/http-native/src/main/java/org/ballerinalang/net/transport/contractimpl/sender/states/http2/ReceivingHeaders.java
+++ b/http-native/src/main/java/org/ballerinalang/net/transport/contractimpl/sender/states/http2/ReceivingHeaders.java
@@ -280,7 +280,7 @@ public class ReceivingHeaders implements SenderState {
         }
         // Create HTTP Carbon Response
         HttpCarbonResponse responseCarbonMsg = new HttpCarbonResponse(httpResponse, new Http2InboundContentListener(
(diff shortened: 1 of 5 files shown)


http-native/src/main/java/org/ballerinalang/net/transport/contractimpl/sender/states/http2/ReceivingHeaders.java
```diff
@@ -203,6 +207,29 @@ private void onResponseHeadersRead(ChannelHandlerContext ctx, OutboundMsgHolder
         }
     }
 
+    private void checkForTrailerHeaders(int streamId, Http2Headers http2Headers, OutboundMsgHolder outboundMsgHolder,
```
ayeshLK, line 210: As per my understanding the responsibility of this method is to process the request if the trailer-header exist. So, better to name it as something like `processHttp2Trailer` etc.

TharmiganK: As I have understood, the method first checks whether there is a trailer-header specified and not. And if trailer is not specified it complete the response with default last http content. Even there is a trailer specified, the method checks whether the respective trailers exists or not and if not it again complete the response with default last content. So the method does not simply process trailer, it checks whether trailer is present and call corresponding methods.

chamil321 (approved):

PR #479, opened 2021-06-21
Handle redundant responses for a single request

chamil321:
## Purpose
```ballerina
    resource function get hello(http:Caller caller) returns string? {
        error? a = caller->respond("Hello"); 
        error? b = caller->respond("Hello"); // case 1
        return "Hello"; // case 2
    }
```
In case 1: Return the error saying the request is already being responded rather than throwing the error.
in case 2: Log the error and return from the execution

Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/1509

## Checklist
- [x] Linked to an issue
- [ ] Updated the changelog
- [x] Added tests

### DIFF
--- a/http-native/src/main/java/org/ballerinalang/net/http/HttpUtil.java
+++ b/http-native/src/main/java/org/ballerinalang/net/http/HttpUtil.java
@@ -216,7 +216,7 @@ public class HttpUtil {
     private static final String METHOD_ACCESSED = "isMethodAccessed";
     private static final String IO_EXCEPTION_OCCURRED = "I/O exception occurred";
     private static final String CHUNKING_CONFIG = "chunking_config";
-    private static final String ILLEGAL_FUNCTION_INVOKED = "illegal function invocation";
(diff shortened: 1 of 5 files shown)


http-native/src/main/java/org/ballerinalang/net/http/HttpUtil.java
```diff
@@ -216,7 +216,7 @@
     private static final String METHOD_ACCESSED = "isMethodAccessed";
     private static final String IO_EXCEPTION_OCCURRED = "I/O exception occurred";
     private static final String CHUNKING_CONFIG = "chunking_config";
-    private static final String ILLEGAL_FUNCTION_INVOKED = "illegal function invocation";
+    private static final String ILLEGAL_FUNCTION_INVOKED = "illegal respond: request has already been responded";
```
shafreenAnfar, line 219: How about something like ?

`illegal respond: response has already been sent`

shafreenAnfar (approved):

PR #528, opened 2021-07-14
Enable HTTP Trace and Access Log

TharmiganK:
## Purpose

Move HTTP trace and access log implementation from runtime and enable trace and access log support inside HTTP module

Fixes [`ballerina-platform/ballerina-standard-library/issues/1073`](https://github.com/ballerina-platform/ballerina-standard-library/issues/1073)

## Examples

The HTTP access logs and trace logs are **disabled as default**. To enable, the configurations can be set by the following `config.toml` file:
```toml
[ballerina.http.traceLogConfig]
# Enable printing trace logs in console
console = true              # Default is false
# Specify the file path to save the trace logs  
path = "testTraceLog.txt"   # Optional
# Specify the hostname and port of a socket service to publish the trace logs
host = "localhost"          # Optional
port = 8080                 # Optional

[ballerina.http.accessLogConfig]
# Enable printing access logs in console
console = true              # Default is false
# Specify the file path to save the access logs  
path = "testAccessLog.txt"  # Optional
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests

### DIFF
--- a/ballerina/http_log_manager.bal
+++ b/ballerina/http_log_manager.bal
@@ -16,11 +16,25 @@
 
 import ballerina/jballerina.java;
 
-function init() {
-    setModule();
-    initializeHttpLogs();
+public type TraceLogConfiguration record {|
+    boolean consoleEnabled = false;
+    string path = "";
+    string host = "";
+    int port = 0;
+|};
+
+public type AccessLogConfiguration record {|
+    boolean consoleEnabled = false;
+    string path = "";
+|};
+
+isolated function initializeHttpLogs() {
+    TraceLogConfiguration & readonly traceLogConfig = {};
+    AccessLogConfiguration & readonly accessLogConfig = {};
+    handle httpLogManager = newHttpLogManager(traceLogConfig, accessLogConfig);
 }
 
-function setModule() = @java:Method {
-    'class: "io.ballerina.stdlib.http.api.nativeimpl.ModuleUtils"
+isolated function newHttpLogManager(TraceLogConfiguration traceLogConfig, AccessLogConfiguration accessLogConfig)
+returns handle = @java:Constructor {
+    'class: "io.ballerina.stdlib.http.api.logging.HttpLogManager"
 } external;
--- a/ballerina/http_log_manager.bal
+++ b/ballerina/http_log_manager.bal
@@ -16,6 +16,12 @@
 
 import ballerina/jballerina.java;
 
+# Represents HTTP trace log configuration.
+#
+# + consoleEnabled - Boolean value to enable or disable console trace logs
+# + path - File path to store trace logs
+# + host - Socket hostname to publish the trace logs
+# + port - Socket port to publish the trace logs
 public type TraceLogConfiguration record {|
     boolean consoleEnabled = false;
     string path = "";
@@ -23,6 +29,10 @@ public type TraceLogConfiguration record {|
     int port = 0;
 |};
 
+# Represents HTTP access log configuration.
+#
+# + consoleEnabled - Boolean value to enable or disable console access logs
+# + path - File path to store access logs
 public type AccessLogConfiguration record {|
     boolean consoleEnabled = false;
     string path = "";
--- a/ballerina/http_log_manager.bal
+++ b/ballerina/http_log_manager.bal
@@ -43,11 +43,7 @@ public type AccessLogConfiguration record {|
 configurable TraceLogConfiguration & readonly traceLogConfig = {};
 configurable AccessLogConfiguration & readonly accessLogConfig = {};
 
-isolated function initializeHttpLogs() {
-    _ = initializeHttpLogManager(traceLogConfig, accessLogConfig);
-}
-
-isolated function initializeHttpLogManager(TraceLogConfiguration traceLogConfig, AccessLogConfiguration accessLogConfig)
+isolated function initializeHttpLogs(TraceLogConfiguration traceLogConfig, AccessLogConfiguration accessLogConfig)
 returns handle = @java:Constructor {
     'class: "io.ballerina.stdlib.http.api.logging.HttpLogManager"
 } external;
--- a/ballerina/http_log_manager.bal
+++ b/ballerina/http_log_manager.bal
@@ -38,9 +38,10 @@ public type AccessLogConfiguration record {|
     string path = "";
 |};
 
+configurable TraceLogConfiguration & readonly traceLogConfig = {};
+configurable AccessLogConfiguration & readonly accessLogConfig = {};
+
 isolated function initializeHttpLogs() {
-    TraceLogConfiguration & readonly traceLogConfig = {};
-    AccessLogConfiguration & readonly accessLogConfig = {};
     handle httpLogManager = newHttpLogManager(traceLogConfig, accessLogConfig);
 }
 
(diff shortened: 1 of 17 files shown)


ballerina/http_log_manager.bal
```diff
@@ -0,0 +1,51 @@
...
+
+import ballerina/jballerina.java;
+
+# Represents HTTP trace log configuration.
+#
+# + consoleEnabled - Boolean value to enable or disable console trace logs
+# + path - File path to store trace logs
+# + host - Socket hostname to publish the trace logs
+# + port - Socket port to publish the trace logs
+public type TraceLogConfiguration record {|
```
chamil321, line 25: Let's put a TODO comment here saying that, to make fields optional once runtime team supports that feature

ballerina/http_log_manager.bal
```diff
@@ -0,0 +1,51 @@
...
+    boolean consoleEnabled = false;
```
chamil321, line 26: I think we can change it to just `console`. WDYT?

ballerina/http_log_manager.bal
```diff
@@ -0,0 +1,51 @@
...
+    string host = "";
+    int port = 0;
+|};
+
+# Represents HTTP access log configuration.
+#
+# + consoleEnabled - Boolean value to enable or disable console access logs
+# + path - File path to store access logs
+public type AccessLogConfiguration record {|
+    boolean consoleEnabled = false;
```
chamil321, line 37: This one too?  just `console`. WDYT?

ballerina/http_log_manager.bal
```diff
@@ -0,0 +1,51 @@
...
+|};
+
+configurable TraceLogConfiguration & readonly traceLogConfig = {};
+configurable AccessLogConfiguration & readonly accessLogConfig = {};
+
+isolated function initializeHttpLogs() {
+    handle httpLogManager = newHttpLogManager(traceLogConfig, accessLogConfig);
+}
+
+isolated function newHttpLogManager(TraceLogConfiguration traceLogConfig, AccessLogConfiguration accessLogConfig)
```
chamil321, line 48:
```suggestion
isolated function initializeHttpLogManager(TraceLogConfiguration traceLogConfig, AccessLogConfiguration accessLogConfig)
```

ballerina/http_log_manager.bal
```diff
@@ -0,0 +1,51 @@
...
+public type AccessLogConfiguration record {|
+    boolean consoleEnabled = false;
+    string path = "";
+|};
+
+configurable TraceLogConfiguration & readonly traceLogConfig = {};
+configurable AccessLogConfiguration & readonly accessLogConfig = {};
+
+isolated function initializeHttpLogs() {
+    handle httpLogManager = newHttpLogManager(traceLogConfig, accessLogConfig);
```
chamil321, line 45: Do we use `httpLogManager` anywhere else? If not we can avoid the return value from the function

ballerina/init.bal
```diff
@@ -18,6 +18,7 @@ import ballerina/jballerina.java;
 
 function init() {
     setModule();
+    initializeHttpLogs();
```
chamil321, line 21:
```suggestion
    initializeHttpLogs(traceLogConfig, accessLogConfig);
```
Can we do this and make `initializeHttpLogs` the external function? So we can get rid of `newHttpLogManager`

ballerina/http_log_manager.bal
TharmiganK: Yes I also prefer console since it is a boolean value it is explicitly means whether console trace log should be enabled or not.  I'll change that

ballerina/http_log_manager.bal
TharmiganK: `httpLogManager` is not used elsewhere. But the external function is a Java constructor which returns the object. That's why I assigned it to a value. Anyway to avoid this?

chamil321: We can change the extern function not to return anything.
https://github.com/ballerina-platform/module-ballerina-http/blob/f765707bb4d9a9d29c489df450c2b3b0575d91d3/ballerina/http_request.bal#L660

ballerina/http_log_manager.bal
```diff
@@ -0,0 +1,49 @@
...
+#
+# + console - Boolean value to enable or disable console trace logs
+# + path - File path to store trace logs
+# + host - Socket hostname to publish the trace logs
+# + port - Socket port to publish the trace logs
+public type TraceLogConfiguration record {|
+    boolean console = false;
+    string path = "";
+    string host = "";
+    int port = 0;
```
ayeshLK, line 30: Can't we make these optional fields ?

ballerina/http_log_manager.bal
```diff
@@ -0,0 +1,49 @@
...
+|};
+
+// TODO: Make the fields optional once configurable records support union members
+# Represents HTTP access log configuration.
+#
+# + console - Boolean value to enable or disable console access logs
+# + path - File path to store access logs
+public type AccessLogConfiguration record {|
+    boolean console = false;
+    string path = "";
```
ayeshLK, line 40: ditto

native/src/main/java/io/ballerina/stdlib/http/api/logging/util/LogLevelMapper.java
```diff
@@ -0,0 +1,70 @@
+/*
+ *  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
ayeshLK, line 2: Better to update the copyright statement consistently.

ballerina/http_log_manager.bal
TharmiganK: Currently union member type **()**  is not supported in **configurable** records. Once the support is given we can change these fields to optional.
```ballerina
public type TraceLogConfiguration record {|
    boolean consoleEnabled = false;
    string? path = ();
    string? host = ();
    int? port = ();
|};

configurable TraceLogConfiguration & readonly traceLogConfig = {};
```
The above code returns a compile time error : 
```
configurable variable currently not supported for '(TraceLogConfiguration & readonly)' 
union member type '()' is not supported
```

native/src/main/java/io/ballerina/stdlib/http/api/logging/util/LogLevelMapper.java
TharmiganK: Theses files are just moved from lang repo. That is why I did not change anything. I can change them.

ballerina/http_log_manager.bal
ayeshLK: I think the limitation is there for `nillable` fields. But you could use optional fields.
```ballerina
public type TraceLogConfiguration record {|
    boolean consoleEnabled = false;
    string path?;
    string host?;
    int port?;
|};

configurable TraceLogConfiguration & readonly traceLogConfig = {};
```

chamil321: +1 Shall we do this change and see? If so we can get rid of default values

native/src/main/java/io/ballerina/stdlib/http/api/logging/util/LogLevelMapper.java
chamil321: Yeah, Since those are not newly wrote files, let's keep it as it is.

ayeshLK (approved):

chamil321 (approved):

PR #560, opened 2021-07-30
Introduce anydata as an outbound req/resp data types

chamil321:
Type | Media Type
-- | --
xml | application/xml
string | text/plain
byte[] | application/octet-stream
map\<anydata\>, table<map\<anydata\>>, anydata[] | application/json
int, float, decimal, boolean | application/json

## Examples
```ballerina
type OpenCustomer record {
    string name;
    xml aa;
    byte[] bb;
};

@test:Config {}
public function testSendingOpenRecord() returns error? {
    OpenCustomer customer = { name: "ballerina", aa: xml `<book>Hello World</book>`, bb: "abc".toBytes()};
    http:Response|error resp = outRequestClient->post("/mytest/json", customer);
}
```
## Test matrix

Type | Single | Array | Map | Table
-- | -- | -- | -- |--
string | + | + | + | +
int, float, decimal, boolean, string  | + | + | + | +
xml | + | + | + | +
byte[] | - | + | - | -
Array | +JSON | - | +JSON | +JSON
Map | +JSON | +JSON | - | +JSON
Table | +JSON | +JSON | +JSON | -
Open record | + | + | + | +
Close record  | + | + | + | +

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests

### DIFF
--- a/ballerina-tests/tests/http_client_actions_test.bal
+++ b/ballerina-tests/tests/http_client_actions_test.bal
@@ -105,7 +105,6 @@ service /httpClientActionBE on httpClientActionListenerEP1 {
         checkpanic caller->respond(checkpanic request.getTextPayload());
     }
 
-    // withWhitespacedExpression
     resource function 'default [string id](http:Caller caller, http:Request request) {
         error? res = caller->respond(id);
     }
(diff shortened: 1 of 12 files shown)


chamil321:
This change has a problem with a byte[] and xml being serialised as json and user cannot retrieve the payload as it is. We can bring back this later with serdes implementation.

PR #603, opened 2021-08-24
Fix multiple clients created for same route not using respective config

chamil321:
## Purpose
Currently the Client connection pool key is scheme + host + port. Regardless what the client config properties are, Clients created for same route share one pool. This fix is to differentiate the pools creation and consider configs. 
With this change pools are created per client. 

Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/1727
Related issue : https://github.com/ballerina-platform/ballerina-standard-library/issues/1811

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests


ballerina-tests/tests/http2_mutual_ssl_test.bal
```diff
@@ -123,7 +123,7 @@ http:ClientConfiguration http2MutualSslClientConf2 = {
     http2Settings: { http2PriorKnowledge: true }
 };
 
-@test:Config {}
+//@test:Config {}
```
chamil321, line 126: Disabled temporarily until check this with bashinee. Test needs to be changed.

shafreenAnfar:
The solution looks good to me. But there are a bunch of disabled test cases to be sorted before merging right ?

shafreenAnfar:
Also any reason why we haven't updated the change log ?

chamil321:
> The solution looks good to me. But there are a bunch of disabled test cases to be sorted before merging right ?

Those disable test cases will be enabled via https://github.com/ballerina-platform/module-ballerina-http/pull/605.
Updated changelog

ldclakmal (approved):

PR #655, opened 2021-09-24
Add codeactions to change return type when `http:Caller` is a parameter and to change header param type

IMS94:
## Purpose
$subject

## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [x] Added tests


compiler-plugin-tests/spotbugs-exclude.xml
```diff
@@ -0,0 +1,27 @@
+<!--
+  ~ Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
nadeeshaan: If this is a new file we need to change the licence header's date

nadeeshaan (changes requested):

compiler-plugin-tests/spotbugs-exclude.xml
IMS94: I had changed that already? Can you check now?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpResourceValidator.java
```diff
@@ -699,6 +702,18 @@ private static void updateDiagnostic(SyntaxNodeAnalysisContext ctx, Location loc
         ctx.reportDiagnostic(DiagnosticFactory.createDiagnostic(diagnosticInfo, location));
     }
 
+    private static void updateDiagnostic(SyntaxNodeAnalysisContext ctx, Location location, String argName,
+                                         HttpDiagnosticCodes httpDiagnosticCodes,
+                                         List<DiagnosticProperty<?>> diagnosticProperties) {
+        DiagnosticInfo diagnosticInfo;
+        if (argName == null) {
```
chamil321, line 709: Do we have any invocations with argName being null?

PR #748, opened 2021-11-23
Fix `parseHeader()` to support multiple header values

chamil321:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/2403

## Examples
```ballerina
# Represents the parsed header value details
public type HeaderValue record {|
    # The header value
    string value;
    # Map of header parameters
    map<string> params;
|};

http:HeaderValue[] values = check http:parseHeader("text/plain;level=1;q=0.6, application/xml;level=2");
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests

### DIFF
--- a/ballerina/http_types.bal
+++ b/ballerina/http_types.bal
@@ -119,3 +119,11 @@ public type Links record {|
     # Array of available links
     Link[] links;
 |};
+
+# Represents the parsed header value details
+public type HeaderValue record {|
+    # The header value
+    string value;
+    # Map of header parameters
+    map<string> params;
+|};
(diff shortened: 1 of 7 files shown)


ballerina/http_types.bal
```diff
@@ -119,3 +119,11 @@ public type Links record {|
...
     Link[] links;
 |};
+
+# Represents the parsed header value details
+public type HeaderValue record {|
+    # The header value
+    string value;
+    # Map of header parameters
+    map<string> params;
+|};
```
ayeshLK, line 129: Shall we add standard doc-comments ?
```suggestion
# Represents the parsed header value details
#
# + value - The header value
# + params - Map of header parameters   
public type HeaderValue record {|
    string value;
    map<string> params;
|};
```

shafreenAnfar: I think what @chamil321 has added works too.

ayeshLK (approved):

PR #797, opened 2021-12-22
Improve support for x-form-urlencoded content usage with `http:Client`

ayeshLK:
Fixes [#1705](https://github.com/ballerina-platform/ballerina-standard-library/issues/1705)

## Examples
Initially when sending `x-form-urlencoded` content via `http:Client` developers had to do additional work to generating the text payload. But with this improvement following is possible.
```ballerina
http:Response response = check snowpeak->post("/update", 
    {
        "key1": "value1",
        "key2": "value2"
    }, 
    mediaType=mime:APPLICATION_FORM_URLENCODED
);
```
 
## Checklist
- [X] Linked to an issue
- [X] Updated the changelog
- [X] Added tests


ballerina/http_commons.bal
```diff
@@ -56,16 +58,48 @@ isolated function buildRequest(RequestMessage message) returns Request|ClientErr
...
+    return string`${keyValPair[0]}=${keyValPair[1]}`;
+}
+
+isolated function processJsonContent(anydata message) returns json|ClientError {
+    var result = trap val:toJson(message);
+    if (result is error) {
+        return error InitializingOutboundRequestError("json conversion error: " + result.message(), result);
+    } else {
+        return result;
+    }
```
chamil321, line 100: I think now we can simple return without an `else`

ayeshLK: Fixed with [dbce5f84fed5da34d8e585ca4c4a2631a14a00e2](https://github.com/ballerina-platform/module-ballerina-http/pull/797/commits/dbce5f84fed5da34d8e585ca4c4a2631a14a00e2)

chamil321 (approved):

ballerina/http_commons.bal
```diff
@@ -56,16 +58,47 @@ isolated function buildRequest(RequestMessage message) returns Request|ClientErr
...
         }
     }
     return request;
 }
 
+isolated function processUrlEncodedContent(anydata message) returns string|ClientError {
+    if message is map<string> {
+        string[] messageParams = message.entries().toArray().'map(constructFormEntry);
+        string payload = strings:'join("&", ...messageParams);
+        string|error encodedContent = url:encode(payload, "UTF-8");
```
shafreenAnfar, line 79: I don't think what we have done here is correct. This encodes everything instead of just the value. Therefore the end result looks like this. For the below map.

```ballerina
map<string> message = {foo: "bar", zoo: "foo"};
```

```
foo%3Dbar%26zoo%3Dfoo
```
Instead of,
```
foo=bar&zoo=foo
```

ayeshLK: Fixed with [92bbefa2fa625f533dbb34cc3516a40fe9002f36](https://github.com/ballerina-platform/module-ballerina-http/pull/797/commits/92bbefa2fa625f533dbb34cc3516a40fe9002f36)

shafreenAnfar (approved):

shafreenAnfar:
I think we can improve the implementation a bit.

```ballerina
map<string> pairs = {
  "key1": "value1",
  "key2": "value2"
};
http:Response response = check snowpeak->post("/update", pairs);
```
If the code looks like the above, the user do not need to explicitly mention the media type. We can do a type checking for `map<string>` and default it to `x-www-form-urlencoded`.

This also means the below types also will fall under `x-www-form-urlencoded` instead of `json`

```ballerina
readonly & json j = {key1: "foo", key2: "bar"};
record {|
    string key1;
    string key2;
|} r = {key1: "foo", key2: "bar"};
record {|
    string...;
|} r = {"key1": "foo", "key2": "bar"};
```
80%  of the time `json` would include more than `string` type. Therefore, it shouldn't be a problem.

PR #814, opened 2022-01-06
Fix type cast error issue

kalaiyarasiganeshalingam:
## Purpose
$Subject

Related to https://github.com/ballerina-platform/ballerina-standard-library/issues/1657
## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests


ballerina/caching_http_caching_client.bal
```diff
@@ -269,29 +269,35 @@ isolated function getCachedResponse(HttpCache cache, HttpClient httpClient, Requ
     time:Utc currentT = time:utcNow();
     req.parseCacheControlHeader();
 
-    if (cache.hasKey(getCacheKey(httpMethod, path))) {
-        Response cachedResponse = cache.get(getCacheKey(httpMethod, path));
+    any|error cacheEntry = cache.get(getCacheKey(httpMethod, path));
+    if (cacheEntry !is error && cacheEntry !is null) {
```
ayeshLK, line 273: should this be checking whether `cacheEntry` is `()` or `null` ?

ballerina/caching_http_caching_client.bal
```diff
@@ -269,29 +269,35 @@ isolated function getCachedResponse(HttpCache cache, HttpClient httpClient, Requ
...
+        lock {
+            if (isFreshResponse(cachedResponse, isShared)) {
+                // If the no-cache directive is not set, responses can be served straight from the cache, without
+                // validating with the origin server.
+                if (!isNoCacheSet(reqCache, resCache) && !req.hasHeader(PRAGMA)) {
+                    log:printDebug("Serving a cached fresh response without validating with the origin server");
+                    return cachedResponse;
+                }
+
+                log:printDebug("Serving a cached fresh response after validating with the origin server");
```
ayeshLK, line 298: Shall we use `lock` only for the sections which has identified concurrency issues. I think here the concurrency issue is with `isFreshResponse` method. Hence we could do something like:
```ballerina
boolean isFreshResponse = false;
lock {
  isFreshResponse = isFreshResponse(cachedResponse, isShared);
}
if isFreshResponse {
  // implement logic here
}
```

ayeshLK (dismissed):
LGTM

ballerina/caching_http_caching_client.bal
```diff
@@ -269,20 +269,28 @@ isolated function getCachedResponse(HttpCache cache, HttpClient httpClient, Requ
...
-        setAgeHeader(cachedResponse);
+        lock {
+            setAgeHeader(cachedResponse);
+        }
 
         RequestCacheControl? reqCache = req.cacheControl;
         ResponseCacheControl? resCache = cachedResponse.cacheControl;
 
-        if (isFreshResponse(cachedResponse, isShared)) {
+        boolean isFreshResp = false;
```
chamil321, line 289:
```suggestion
        boolean freshResponse = false;
```

chamil321: Let's maintain the standard of not using the prefix (is/has) for boolean variable naming and use them only for the method naming which returns a boolean type

ballerina/caching_http_caching_client.bal
kalaiyarasiganeshalingam: Earlier, we had a chance to get () or null. Updated the cache module now. So, I removed this check as no need now.

chamil321 (approved):

shafreenAnfar (approved):

PR #821, opened 2022-01-11
Stop adding sourceHandler for async backend calls

TharmiganK:
## Purpose

When there is a call to an async backend, that outbound request should not depend on the SourceHandler and a connection should be created from the global connection pool.

> Fixes [`Connection pool does not get recovered after being exhausted #1356`](https://github.com/ballerina-platform/ballerina-standard-library/issues/1356)

> **Note** : Since this fix is related to the strands which are created to execute client actions, it is not possible to add a test case for this scenario. But the behaviour change after this fix is checked by manual debugging and observed that, for async backend call, the connection is created from the **global pool** and the TargetChannel for this connection is **not** associated with a SourceHandler.

## Example

```ballerina
service Hello on new http:Listener(9090) {
    @http:ResourceConfig {
        path: "/sayHello"
    }
    resource function passthrough(http:Caller caller, http:Request req) {
        // creates a connection from global pool
        future<()> publishedEvent = start publishSecondEvent(); 

        // creates a connection from SourceHandler's targetChannelPool
        var clientResponse = clientEP->forward("/sayHello", req); 
        if (clientResponse is http:Response) {
            var result = caller->respond(clientResponse);
            if (result is error) {
                log:printError("Error sending response", result);
            }
        } else {
            http:Response res = new;
            res.statusCode = 500;
            res.setTextPayload(<string>clientResponse.detail()?.message);
            var result = caller->respond(res);
            if (result is error) {
                log:printError("Error sending response", result);
            }
        }
    }
}

public function publishSecondEvent() {
    http:Request clientRequest = new;
    clientRequest.setTextPayload("Hello World");
    var response = clientEP2->post("/sayHello", clientRequest);
    if (response is http:Response) {
        log:printDebug("Second request sent.");
    } else {
        log:printError(response.reason(), err = response);
    }
}
```

## Checklist
- [x] Linked to an issue
- [ ] <s>Added tests</s>

### DIFF
--- a/ballerina/Ballerina.toml
+++ b/ballerina/Ballerina.toml
@@ -13,7 +13,7 @@ distribution = "2201.0.0"
 path = "../native/build/libs/http-native-2.2.0-SNAPSHOT.jar"
 
 [[platform.java11.dependency]]
-path = "./lib/mime-native-2.2.0-20220110-185500-739a5a4.jar"
+path = "./lib/mime-native-2.2.0-20220111-111400-09ccefa.jar"
 
 [[platform.java11.dependency]]
 path = "./lib/netty-common-4.1.71.Final.jar"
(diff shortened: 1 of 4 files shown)


chamil321 (approved):

shafreenAnfar (approved):
LGTM!

PR #833, opened 2022-01-13
Add compiler validation for payload annotation usage

TharmiganK:
## Purpose

A compiler error is returned when a non entity body resource (`get`, `options` and `head`) has a parameter with payload annotation(`@http:Payload`)

> Fixes [`No validation for get with @http:Payload #2561`](https://github.com/ballerina-platform/ballerina-standard-library/issues/2561)

## Example
```ballerina
import ballerina/http;

service / on new http:Listener(9090) {
    // will return a compiler error saying : 
    // invalid usage of payload annotation for non entity body resource : 'get'. 
    // Use an accessor that supports entity body
    resource function get hello(@http:Payload string request) returns string {
        return request;
    }
}
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests

### DIFF
--- a/compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpDiagnosticCodes.java
+++ b/compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpDiagnosticCodes.java
@@ -74,8 +74,8 @@ public enum HttpDiagnosticCodes {
             "'[string... path]', but found '%s'", ERROR),
     HTTP_128("HTTP_128", "invalid interceptor resource method: expected default resource method: " +
             "'default', but found '%s'", ERROR),
-    HTTP_129("HTTP_129", "invalid payload annotation usage for non entity body resource : '%s'. " +
-            "Use accessor that supports entity", ERROR),
+    HTTP_129("HTTP_129", "invalid usage of payload annotation for a non entity body resource : '%s'. " +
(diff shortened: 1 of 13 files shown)


chamil321 (approved):

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpDiagnosticCodes.java
```diff
@@ -74,6 +74,7 @@
             "'[string... path]', but found '%s'", ERROR),
     HTTP_128("HTTP_128", "invalid interceptor resource method: expected default resource method: " +
             "'default', but found '%s'", ERROR),
+    HTTP_129("HTTP_129", "invalid payload annotation usage for non entity body resource : '%s'", ERROR),
```
chamil321, line 77:
```suggestion
    HTTP_129("HTTP_129", "invalid payload annotation usage for non entity body resource : '%s'. Use accessor that supports entity", ERROR),
```

TharmiganK: Fixed via [`80d0e55`](https://github.com/ballerina-platform/module-ballerina-http/pull/833/commits/80d0e55b34b3d2e3ea493af3c12104e382132979)

chamil321 (approved):

PR #837, opened 2022-01-18
Remove backPressureObservable listener after failure

TharmiganK:
## Purpose

When the channel is not writable but active, it acquires a semaphore and waits until it get notified by `onWritable()`. So when there is a failure the BackPressureObservable listener should be removed in order to release the semaphore through `onWritable()`. 

## Examples

N/A

## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests


native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/DefaultHttpResponseFuture.java
```diff
@@ -118,6 +119,12 @@ public void notifyHttpListener(HttpCarbonMessage httpCarbonMessage) {
     public void notifyHttpListener(Throwable throwable) {
         responseLock.lock();
         try {
+            if (outboundMsgHolder != null) {
```
shafreenAnfar, line 122: Do we actually use this constructor if not we don't need the above null check.

https://github.com/ballerina-platform/module-ballerina-http/pull/837/files#diff-3f887598e6e7155d27926134275f68ae74193fc70bc11a3935ee8319206a608cR73

TharmiganK: Yes, we are using the constructor with `null` outboundMsgHolder, in `notifyListenerAndGetErrorResponseFuture` method and in `HttpCarbonMessage` class.

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/DefaultHttpResponseFuture.java
```diff
@@ -118,6 +120,11 @@ public void notifyHttpListener(HttpCarbonMessage httpCarbonMessage) {
     public void notifyHttpListener(Throwable throwable) {
         responseLock.lock();
         try {
+            if (backPressureHandler != null) {
+                backPressureHandler.getBackPressureObservable().removeListener();
+            } else if (outboundMsgHolder != null) {
+                outboundMsgHolder.getBackPressureObservable().removeListener();
+            }
```
shafreenAnfar, line 127: else we can print a warning saying no BackPressureObservable found.

shafreenAnfar: Also, lets add a comment explaining why we have two of them.

TharmiganK: Resolved in [`557b8e3 `](https://github.com/ballerina-platform/module-ballerina-http/pull/837/commits/557b8e394a65a39f247b739acf498a71fd2b462e)

shafreenAnfar (dismissed):

shafreenAnfar (approved):

PR #839, opened 2022-01-18
Update default content-type for `map<string>` as `application/x-www-form-urlencoded`

ayeshLK:
Fixes [#2589](https://github.com/ballerina-platform/ballerina-standard-library/issues/2589)
Related to [#1705](https://github.com/ballerina-platform/ballerina-standard-library/issues/1705#issuecomment-1014781301)

## Checklist
- [X] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests


ballerina-tests/tests/http_client_data_binding.bal
```diff
@@ -445,7 +445,7 @@ service /backend on clientDBBackendListener {
 
     resource function get jsontype() returns http:InternalServerError {
         json j = {id:"hello"};
-        return {body: j};
+        return {body: j, mediaType: mime:APPLICATION_JSON};
```
shafreenAnfar, line 448: Do we need the mediaType here ?

ballerina-tests/tests/http_client_url_encoded_content_test.bal
```diff
@@ -46,6 +46,17 @@ isolated function testUrlContentWithPost() returns error? {
...
 }
 
+@test:Config {
+    groups: ["urlEncodedContent"]
+}
+isolated function testUrlContentInline() returns error? {
+    string response = check clientUrlEncodedTestClient->post("", {
+        "key1": "value1",
+        "key2": "value2"
+    }, mediaType = mime:APPLICATION_FORM_URLENCODED);
```
shafreenAnfar, line 56: Let's properly format this.

shafreenAnfar:
```ballerina
string response = check clientUrlEncodedTestClient->post("", 
  {
    "key1": "value1",
    "key2": "value2"
  }, 
  mediaType = mime:APPLICATION_FORM_URLENCODED
);
```

ayeshLK: Fixed with [e68158206c5ae3747c86e5af80f533ea2b987e18](https://github.com/ballerina-platform/module-ballerina-http/pull/839/commits/e68158206c5ae3747c86e5af80f533ea2b987e18)

ballerina-tests/tests/http_client_data_binding.bal
ayeshLK: Fixed with [b927caec215436c0c1174ccf51d564e55c8ea4cd](https://github.com/ballerina-platform/module-ballerina-http/pull/839/commits/b927caec215436c0c1174ccf51d564e55c8ea4cd)

ballerina/http_commons.bal
```diff
@@ -58,40 +58,50 @@ isolated function buildRequest(RequestMessage message, string? mediaType) return
         request.setByteStream(message);
     } else if (message is mime:Entity[]) {
         request.setBodyParts(message);
+    } else if message is map<string> {
+        match mediaType {
+            mime:APPLICATION_JSON => {
+                json payload = check processJsonContent(message);
+                request.setJsonPayload(payload);
+            }
+            _ => {
```
chamil321, line 67: What happens if user sets the media type as something other than `application/x-www-form-urlencoded` after passing a `map<string>`. According to that logic, we still encode and set as text payload. I think we should check the mediatype is null or `application/x-www-form-urlencoded` first. Then we should reat as `application/x-www-form-urlencoded`. For other media types, we parse them according to the media type or give an error. @shafreenAnfar WDYT?

ballerina/http_connection.bal
```diff
@@ -258,6 +260,15 @@ isolated function setPayload(anydata payload, Response response, string? mediaTy
         if setETag {
             response.setETag(payload);
         }
+    } else if payload is map<string> {
+        match mediaType {
+            mime:APPLICATION_JSON => {
+                setJsonPayload(response, payload, setETag);
+            }
+            _ => {
```
chamil321, line 268: Here also the same concern applies.

ballerina/http_connection.bal
```diff
@@ -266,8 +277,9 @@ isolated function setPayload(anydata payload, Response response, string? mediaTy
 isolated function processAnydata(Response response, anydata payload, string? mediaType = (), boolean setETag = false) {
     match mediaType {
         mime:APPLICATION_FORM_URLENCODED => {
-            if payload is map<string> {
-                string|error result = retrieveUrlEncodedData(payload);
+            map<string>|error pairs = checkpanic val:cloneWithType(payload);
```
chamil321, line 280: Here once you do the `checkpanic`, there wont be any error left as a return type right?

ballerina/http_commons.bal
shafreenAnfar: I think same level of improvement can come later. The same logic applies for the other scenarios as well. For instance, if you send `bye[]` along with media type `text/plain`, it works without giving any error. If I am not mistaken. 

I believe, our data binding implementations needs to be changed as https://github.com/ballerina-platform/ballerina-standard-library/issues/2530.  We need to do this soon after SL GA.

ballerina/http_connection.bal
```diff
@@ -139,6 +140,7 @@ public isolated client class Caller {
         } else if message is error {
             if message is ApplicationResponseError {
                 InternalServerError err = {
+                    mediaType: returnMediaType is ()? mime:APPLICATION_JSON: returnMediaType,
```
shafreenAnfar, line 143: @chamil321 are you okay with this change ?

chamil321: We pass the `returnMediaType` to the `createStatusCodeResponse()` method and set the content type according to below order. So if we override it, probably `mediaType` field will get high priority when user set it or not(I mean when () is passed.)
```
    // Update content-type header according to the priority. (Highest to lowest)
    // 1. MediaType field in response record
    // 2. Payload annotation mediaType value
    // 3. The content type header included in headers field
    // 4. Default content type related to payload
```

chamil321: Since we are setting the header to Error response, I think it's fine.

ballerina/http_connection.bal
```diff
@@ -19,6 +19,7 @@ import ballerina/lang.value as val;
 import ballerina/lang.'string as strings;
 import ballerina/url;
 import ballerina/mime;
+import ballerina/log;
```
chamil321, line 22:
```suggestion
```

ballerina/http_connection.bal
chamil321, line 280:
```suggestion
            map<string>|error pairs = val:cloneWithType(payload);
```

ballerina/http_connection.bal
chamil321: Correction on the above comment. 
I think this change is wrong. With this change all every data type gets only the `mime:APPLICATION_JSON` as it's content type whenever the return type is not mentioned

ballerina/http_connection.bal
```diff
@@ -138,11 +138,12 @@ public isolated client class Caller {
             }
         } else if message is error {
             if message is ApplicationResponseError {
+                string? retrievedMediaType = retrieveMediaTypeForErrorResponse(returnMediaType, message.detail().get(CONTENT_TYPE));
```
chamil321: error detail does not contain a field called CONTENT_TYPE. You need to access the headers

ballerina/http_connection.bal
```diff
@@ -184,6 +185,18 @@ public isolated client class Caller {
     }
 }
 
+isolated function retrieveMediaTypeForApplicationError(string? annotatedMediaType, anydata retrievedMediaType) returns string? {
```
shafreenAnfar, line 188: Ins't retrievedMediaType type should be `string?` ?

ayeshLK: Fixed with [28f2606e58a20d925a1569c3070c9fe670884718](https://github.com/ballerina-platform/module-ballerina-http/pull/839/commits/28f2606e58a20d925a1569c3070c9fe670884718)

ballerina/http_connection.bal
ayeshLK: Fixed with [28f2606e58a20d925a1569c3070c9fe670884718](https://github.com/ballerina-platform/module-ballerina-http/pull/839/commits/28f2606e58a20d925a1569c3070c9fe670884718)

ballerina/http_connection.bal
```diff
@@ -184,6 +186,24 @@ public isolated client class Caller {
...
+    if annotatedMediaType is string {
+        return annotatedMediaType;
+    } else {
+        if !headers.hasKey(CONTENT_TYPE) {
+            return;
+        } 
+        string[]|string mediaType = headers.get(CONTENT_TYPE);
+        if mediaType is string {
+            return mediaType;
+        } else {
```
shafreenAnfar, line 199: There can't be an array of media types right ?

ayeshLK: Fixed with [9a0b4f39dab43eda54e01b64619dd0baaa658eaf](https://github.com/ballerina-platform/module-ballerina-http/pull/839/commits/9a0b4f39dab43eda54e01b64619dd0baaa658eaf)

chamil321 (approved):

chamil321 (approved):

chamil321 (approved):

ballerina/http_connection.bal
ayeshLK: When setting `content-type` header to error-details it is set as a `string[]`. Hence reverted the fix.

shafreenAnfar (approved):

shafreenAnfar (approved):

PR #852, opened 2022-01-26
Fix BackPressureObservable issue for outbound response path

TharmiganK:
## Purpose

> $Subject

> Fixes [`BackPressureObservable listener should be removed after failure #2607`](https://github.com/ballerina-platform/ballerina-standard-library/issues/2607)

## Examples

N/A

## Checklist
- [x] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests


native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/DefaultHttpResponseFuture.java
```diff
@@ -47,7 +47,8 @@
     private HttpConnectorListener pushPromiseListener;
     private ConcurrentHashMap<Integer, HttpConnectorListener> pushResponseListeners;
     private ConcurrentHashMap<Integer, Throwable> pushResponseListenerErrors;
-    private BackPressureHandler backPressureHandler;
+    private BackPressureHandler requestBackPressureHandler;
+    private BackPressureHandler responseBackPressureHandler;
```
shafreenAnfar, line 51: At any given time, there could be only one of these two. Because the same CarbonMessage can't be both outbound request and outbound response. Therefore, can't we just have one field as we did earlier ?

TharmiganK: Fixed via [`9b3e185`](https://github.com/ballerina-platform/module-ballerina-http/pull/852/commits/9b3e18564129f40f3c02abd9d85db3910db543d0)

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/sender/channel/TargetChannel.java
```diff
@@ -82,8 +83,11 @@ public void configTargetHandler(HttpCarbonMessage httpCarbonMessage, HttpRespons
         targetHandler.setTargetChannel(this);
     }
 
-    public BackPressureHandler getBackPressureHandler() {
-        return Util.getBackPressureHandler(targetHandler.getContext());
+    public BackPressureObservable getBackPressureObservable() {
+        if (Util.getBackPressureHandler(targetHandler.getContext()) != null) {
+            return Util.getBackPressureHandler(targetHandler.getContext()).getBackPressureObservable();
+        }
+        return null;
```
shafreenAnfar, line 90: Instead of returning `null`, I think we should throw an exception.

TharmiganK: In this part, we check whether the `backPressureHandler` is empty or not to extract the `backPressureObservable`. Since, this is used for HTTP1.1 and HTTP2, we will not have a `backPressureHandler` for HTTP2 (For HTTP2, we get the `backPressureObservable` from `outboundMsgHolder`). That's why we return `null`, and we only update `backPressureHandler` in the `DefaultHttpResponseFuture` if it is not null. Throwing an exception in this point causes test failures.

chamil321 (approved):

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/sender/channel/TargetChannel.java
shafreenAnfar: Returning `null` from an API is not good. Because by looking at the API one cannot decide all the possible outcomes.  But in this case, we can tolerate it.

shafreenAnfar (approved):

PR #871, opened 2022-02-06
Extend doc comments for http:Response

Gubaer:
## Purpose
Improve doc comments in `http:Response`.

The method `getTextPayload()` returns a `http:NoContentError` if the payload is empty, not an empty string. This patch  extends the doc comment to make this clear. 

For consistency reasons, it extends the doc comments for `getXmlPayload()` and `getJsonPayload()` too.

### DIFF
--- a/ballerina/http_response.bal
+++ b/ballerina/http_response.bal
@@ -194,7 +194,11 @@ public class Response {
         return contentTypeHeaderValue;
     }
 
-    # Extract `json` payload from the response. If the content type is not JSON, an `http:ClientError` is returned.
+    # Extract `json` payload from the response.
+    #
+    # If the payload is empty, a `http:NoContentError` is returned.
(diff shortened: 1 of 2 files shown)


ballerina/http_response.bal
```diff
@@ -218,6 +222,8 @@ public class Response {
 
     # Extracts `xml` payload from the response.
     #
+    # If the payload is empty, a `http:NoContentError` is returned.
+    #
```
chamil321, line 226:
```suggestion
    # Extracts `xml` payload from the response. For an empty payload, `http:NoContentError` is returned.
    #
```

chamil321:
@Gubaer Thanks for the contribution. Let's update the http_request.bal also with the same fix.

chamil321 (approved):

PR #892, opened 2022-02-15
Refactor the code base and remove anti-patterns

chamil321:
## Checklist
- [x] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests


ballerina-tests/tests/configuration_request_limit_config_test.bal
```diff
@@ -109,7 +109,7 @@ service /requestPayloadLimit on lowPayloadLimitEP {
 function testValidUrlLength() {
     http:Client limitClient = checkpanic new("http://localhost:" + requestLimitsTestPort1.toString());
```
ayeshLK, line 110: Lets remove `checkpanic` usage and use `check` instead. Check for other places as well.
```suggestion
function testValidUrlLength() returns error? {
    http:Client limitClient = check new("http://localhost:" + requestLimitsTestPort1.toString());
```

ballerina-tests/tests/configuration_request_limit_config_test.bal
```diff
@@ -109,7 +109,7 @@ service /requestPayloadLimit on lowPayloadLimitEP {
...
     http:Response|error response = limitClient->get("/requestUriLimit/validUrl");
```
ayeshLK, line 111: We could do as follows:
```ballerina
http:Response response = check limitClient->get("/requestUriLimit/validUrl");
// do other validations
```

chamil321: Since this is not an anti-pattern, kept the test cases as it is to handle errors

ballerina-tests/tests/configuration_accept_encoding_test.bal
```diff
@@ -33,17 +33,17 @@ final http:Client acceptEncodingDisableEP = check new("http://localhost:" + acce
 
 service /hello on acceptEncodingListenerEP {
 
-    resource function 'default .(http:Caller caller, http:Request req) {
+    resource function 'default .(http:Caller caller, http:Request req) returns error? {
         http:Response res = new;
         json payload = {};
         boolean hasHeader = req.hasHeader(ACCEPT_ENCODING);
         if (hasHeader) {
```
ayeshLK, line 40:
```suggestion
        if hasHeader {
```

ballerina-tests/tests/configuration_accept_encoding_test.bal
```diff
@@ -82,7 +78,7 @@ function testAcceptEncodingAuto() {
     http:Request req = new;
     req.setTextPayload("accept encoding test");
     http:Response|error response = acceptEncodingAutoEP->post("/", req);
```
ayeshLK, line 80: Shall we do as follows:
```ballerina
    http:Response response = check acceptEncodingAutoEP->post("/", req);
    // do the validations
```
Check for other places as well.

ballerina-tests/tests/http2_100_continue_test.bal
```diff
@@ -50,30 +50,30 @@ service /helloWorld on new http:Listener(9127, {httpVersion: "2.0"}) {
...
 
 @test:Config {}
-public function testUnexpected100ContinueResponse() {
-    http:Client clientEP = checkpanic new("http://localhost:9128");
+public function testUnexpected100ContinueResponse() returns error? {
+    http:Client clientEP = check new("http://localhost:9128");
     http:Response|error resp = clientEP->get("/continueService/initial");
-    if (resp is http:Response) {
+    if resp is http:Response {
         var payload = resp.getTextPayload();
```
ayeshLK, line 75: Shall we do like this:
```ballerina
        string payload = check resp.getTextPayload();
        // do validations
```

ballerina-tests/tests/http2_client_upgrade_with_large_payload_test.bal
```diff
@@ -15,25 +15,21 @@
...
-        if (jsonPayload is json) {
-            response.setPayload(jsonPayload);
-            error? result = caller->respond(response);
-        } else {
-            log:printError("Error getting the json payload");
-        }
+        json jsonPayload = check request.getJsonPayload();
+        response.setPayload(jsonPayload);
+        check caller->respond(response);
+        return;
```
ayeshLK, line 29: I think we can ommit `return;` here.
```suggestion
        return;
```

ballerina-tests/tests/configuration_accept_encoding_test.bal
chamil321: > Shall we do as follows:
> 
> ```ballerina
>     http:Response response = check acceptEncodingAutoEP->post("/", req);
>     // do the validations
> ```
> 
> Check for other places as well.

Check my comment on your above suggestion. Since this is not an anti-pattern, kept the test cases as it is to handle errors

TharmiganK (approved):

PR #903, opened 2022-02-24
Pass empty map values instead of a null for the error detail

chamil321:
## Purpose
$subject

## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests


chamil321:
Respective lang change has been reverted. Hence closing

PR #1006, opened 2022-05-04
Add `openApiDefinition` field to `@http:ServiceConfig` annotation

ayeshLK:
In the initial design we planned to add the `openApiDefinition` to the `@openapi:ServiceInfo` annotation. But following concern arises when we use that approach:
- If the developer has not specified any open-api related functionality, if we are to add `@openapi:ServiceInfo` using code-modifier then we have to add the `ballerina/openapi` import statement as well.

Since, `openApiDefinition` field is used by ballerina HTTP package itself we could add that particular field to the `@http:ServiceConfig` annotation itself. 

Related to [#2472](https://github.com/ballerina-platform/ballerina-standard-library/issues/2472)

## Examples
N/A

## Checklist
- [X] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests
- [x] Updated the spec

### DIFF
--- a/ballerina/http_annotation.bal
+++ b/ballerina/http_annotation.bal
@@ -24,6 +24,7 @@
 # + mediaTypeSubtypePrefix - Service specific media-type subtype prefix
 # + treatNilableAsOptional - Treat Nilable parameters as optional
 # + interceptors - An array of interceptor services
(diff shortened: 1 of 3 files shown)


TharmiganK (approved):

shafreenAnfar:
LGTM!

PR #1009, opened 2022-05-06
Add union type support for service data binding

chamil321:
## Purpose
The payload parameter type can be a union of `anydata`. Based on the media type, the potential binding
type is decided. For example, if the union is defined as `json|xml` and the media type is related to `json`, 
the deserialization and the binding will proceed according to type `json`. 
If the given types of the union are not compatible with the media type, an error is returned.

If any of the type is union with `()`(i.e `string?`), then in the absence of the payload, `()` will be assigned as 
the value without being responded by a `BAD REQUEST` response.

Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/2701

## Examples
```ballerina
resource function post checkUnionTypesWithXmlJson(@http:Payload json|xml payload) returns json|xml {
    if payload is json {
        return {result: "It's json"};
    } else {
        return payload;
    }
}

resource function post checkUnionTypesWithStringNil(@http:Payload string? payload) returns json {
    if payload is string {
        return {result: payload};
    } else {
        return {result: "It's nil"};
    }
}
```    

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec


native/src/main/java/io/ballerina/stdlib/http/api/service/signature/PayloadParam.java
```diff
@@ -89,21 +92,35 @@ private void validatePayloadParam(Type parameterType) {
...
             }
-            this.readonly = true;
             for (Type type : memberTypes) {
                 if (type.getTag() == TypeTags.READONLY_TAG) {
+                    this.readonly = true;
                     continue;
                 }
-                this.type = type;
-                break;
+//                if (type.getTag() == TypeTags.UNION_TAG) {
```
TharmiganK, line 103: Do we need these commented lines?

TharmiganK (approved):

PR #1012, opened 2022-05-06
Add union type support for client data binding

chamil321:
## Purpose
The client action return type can be a union of `anydata`. Based on the media type, the potential binding
type is decided. For example, if the union is defined as `json|xml` and the media type is related to `json`,
the deserialization and the binding will proceed according to the type `json`. But if the media type is related to `xml`
the process will happen according to the type `xml`.
If the given types of the union are not compatible with the media type, an error is returned.

If any of the type is union with `()`(i.e `string?`), then in the absence of the payload, `()` will be assigned as 
the value without being responded by a `BAD REQUEST` response.

Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/2883

## Examples
```ballerina
xml|json response = check clientDBBackendClient->get("/anydataTest/jsonType");
string|xml person = check clientDBBackendClient->get("/anydataTest/getFormData");
Person|Stock response = check clientDBBackendClient->get("/anydataTest/recordType");
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec

### DIFF
--- a/ballerina-tests/tests/http_client_data_binding_anydata.bal
+++ b/ballerina-tests/tests/http_client_data_binding_anydata.bal
@@ -673,6 +673,46 @@ function testUnionOfDatabindingForRecordWithPersonNStock() returns error? {
 }
 
 @test:Config {}
+function testUnionOfDatabindingForMoreTypesWithString() returns error? {
+    xml|json|string|byte[] response = check clientDBBackendClient->get("/anydataTest/stringType");
+    if response is string {
+        test:assertEquals(response, "hello", msg = "Found unexpected output");
+    } else {
+        test:assertFail(msg = "Found unexpected output type");
+    }
+}
+
+@test:Config {}
+function testUnionOfDatabindingForMoreTypesWithJson() returns error? {
+    xml|json|string|byte[] response = check clientDBBackendClient->get("/anydataTest/jsonType");
+    if response is json {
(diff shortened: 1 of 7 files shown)


ballerina-tests/tests/http_client_data_binding_anydata.bal
```diff
@@ -545,3 +588,128 @@ function testXmlArrDatabindingByType() {
...
+function testNilableUnionOfDatabindingForOctetBlob() returns error? {
+    xml|byte[]? response = check clientDBBackendClient->get("/anydataTest/getNoContentBlob");
+    test:assertTrue(response is (), msg = "Found unexpected output");
+}
+
+@test:Config {}
+function testNilableUnionOfDatabindingForJson() returns error? {
+    xml|json? response = check clientDBBackendClient->get("/anydataTest/getNoContentJson");
+    test:assertTrue(response is (), msg = "Found unexpected output");
+}
```
shafreenAnfar, line 715: Do we have a test-case to cover union types more than two ?

chamil321: Added via 9a47153f00311d6798ec1db66fbdfbb31f051e66

shafreenAnfar (approved):

PR #1054, opened 2022-05-27
Allow builtin sub-types for client/resource data binding and resource return types

chamil321:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/2924

## Checklist
- [x] Linked to an issue
- [ ] Updated the changelog
- [x] Added tests
- [ ] Updated the spec


compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
```diff
@@ -209,10 +207,16 @@ public static TypeDescKind retrieveEffectiveTypeDesc(TypeSymbol descriptor) {
...
-    private static boolean isBasicTypeDesc(TypeDescKind kind) {
-        return kind == TypeDescKind.STRING || kind == TypeDescKind.INT || kind == TypeDescKind.FLOAT ||
-                kind == TypeDescKind.DECIMAL || kind == TypeDescKind.BOOLEAN || kind == TypeDescKind.JSON ||
-                kind == TypeDescKind.XML || kind == TypeDescKind.RECORD || kind == TypeDescKind.BYTE;
+    public static boolean isAnyDataType(TypeDescKind kind) {
+        return kind == TypeDescKind.BOOLEAN || kind == TypeDescKind.INT || kind == TypeDescKind.FLOAT ||
+                kind == TypeDescKind.DECIMAL || kind == TypeDescKind.STRING || kind == TypeDescKind.XML ||
+                kind == TypeDescKind.JSON || kind == TypeDescKind.RECORD || kind == TypeDescKind.ANYDATA ||
+                kind == TypeDescKind.NIL || kind == TypeDescKind.BYTE || kind == TypeDescKind.STRING_CHAR ||
+                kind == TypeDescKind.XML_ELEMENT || kind == TypeDescKind.XML_COMMENT ||
```
TharmiganK, line 215: I don't think having all the inbuilt sub-types is a good approach, since we also allow user to create subtypes from basic types. Consider the following example : 
```ballerina
import ballerina/http;

type newXml xml;

service on new http:Listener(9090) {
    resource function get .() returns newXml {
        return xml`<name>Tharmigan</name>`;
    }
}
```

I think the above example will fail with this change since the `newXml` is not considered in the subtypes. Ideally, these type kinds should fall under `TYPE_REFERENCE` and we should allow this type if its effective type is allowable basic type. WDYT?

chamil321: I'll add a test case to cover above case

chamil321: @TharmiganK As you said, the custom subtypes which created by the user are resolved to their basic effective type. In this PR, the built-in subtypes are added which we had missed earlier.

```
type newXml xml;
```
Above is resolved to the basic type of `xml`, but
```
```
type newXmlElement xml:Element;
```
is not resolved to the basic type of `xml`. It resolved to the builtin subtype of `xml:Element`.

Hence we need to add subtype also to the condition.

File: compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java [hunk 2]
```diff
@@ -209,10 +207,16 @@ public static TypeDescKind retrieveEffectiveTypeDesc(TypeSymbol descriptor) {
         return typeDescKind;
     }
 
-    private static boolean isBasicTypeDesc(TypeDescKind kind) {
-        return kind == TypeDescKind.STRING || kind == TypeDescKind.INT || kind == TypeDescKind.FLOAT ||
-                kind == TypeDescKind.DECIMAL || kind == TypeDescKind.BOOLEAN || kind == TypeDescKind.JSON ||
-                kind == TypeDescKind.XML || kind == TypeDescKind.RECORD || kind == TypeDescKind.BYTE;
+    public static boolean isAnyDataType(TypeDescKind kind) {
+        return kind == TypeDescKind.BOOLEAN || kind == TypeDescKind.INT || kind == TypeDescKind.FLOAT ||
```
ayeshLK, line 211: Rather than having a huge boolean condition can't we have a list of all the types included in `anydata` and verify whether the given type is available in that list ? I feel that it would be much more readable than this approach.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
TharmiganK: Yes, that is correct, but `xml:Element` is also a type of `xml` right? So even without adding `xml:Element` to the condition I think this will get resolve in two steps:
1. Will get the type as `newXmlElement` -> go to `TYPE_REFERENCE` case -> get the effective type as `xml:Element` -> will call `validateReturnType`
2. Will get the type as `xml:Element` -> go to `TYPE_REFERENCE` case (when the subtype is not added in the condition` -> get the effective type as `xml` -> will call `validateReturnType`
Finally now we have our basic type : `xml`
Am I correct 🤔

native/src/main/java/io/ballerina/stdlib/http/api/service/signature/builder/XmlPayloadBuilder.java
```diff
@@ -40,7 +40,11 @@ public XmlPayloadBuilder(Type payloadType) {
 
     @Override
     public Object getValue(BObject entity, boolean readonly) {
-        if (isSubtypeOfAllowedType(payloadType, TypeTags.XML_TAG)) {
+        if (isSubtypeOfAllowedType(payloadType, TypeTags.XML_TAG) ||
+                isSubtypeOfAllowedType(payloadType, TypeTags.XML_ELEMENT_TAG) ||
+                isSubtypeOfAllowedType(payloadType, TypeTags.XML_COMMENT_TAG) ||
+                isSubtypeOfAllowedType(payloadType, TypeTags.XML_PI_TAG) ||
+                isSubtypeOfAllowedType(payloadType, TypeTags.XML_TEXT_TAG)) {
```
ayeshLK, line 47:
```suggestion
        List<TypeTags> allowedTypes = Arrays.asList(TypeTags.XML_TAG, TypeTags.XML_ELEMENT_TAG, TypeTags.XML_COMMENT_TAG, TypeTags.XML_PI_TAG, TypeTags.XML_TEXT_TAG);
        boolean payloadTypeAllowed = allowedTypes.stream().anyMatch(at -> isSubtypeOfAllowedType(payloadType, at)); 
        if (payloadTypeAllowed) {
```

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
TharmiganK: In additon to the compiler tests, can we also add a ballerina test case to make sure the `PayloadBuilder` works with custom types

chamil321: > Yes, that is correct, but `xml:Element` is also a type of `xml` right? So even without adding `xml:Element` to the condition I think this will get resolve in two steps:
> 
> 1. Will get the type as `newXmlElement` -> go to `TYPE_REFERENCE` case -> get the effective type as `xml:Element` -> will call `validateReturnType`
> 2. Will get the type as `xml:Element` -> go to `TYPE_REFERENCE` case (when the subtype is not added in the condition -> get the effective type as `xml` -> will call `validateReturnType`
>    Finally now we have our basic type : `xml`
> 
> Am I correct 🤔
> 
> If so we don't need to add the subtypes to the condition

No, As I mentioned above, `xml:Element` is not further resolved to `xml`.

TharmiganK: Got it 👍

TharmiganK (approved):

chamil321:
Related PR https://github.com/ballerina-platform/module-ballerina-http/pull/1141
Related PR https://github.com/ballerina-platform/module-ballerina-http/pull/1140

PR #1124, opened 2022-07-06
Fix payload binding failure in HTTP2

TharmiganK:
> Fixes : [`Payload binding in resource method is not working with HTTP2 #3020`](https://github.com/ballerina-platform/ballerina-standard-library/issues/3020)

## Examples

N/A

## Checklist
- [x] Linked to an issue
- [x] Added tests
- [ ] <s>Updated the spec</s>

### DIFF
--- a/native/src/main/java/io/ballerina/stdlib/http/api/BallerinaHTTPConnectorListener.java
+++ b/native/src/main/java/io/ballerina/stdlib/http/api/BallerinaHTTPConnectorListener.java
@@ -139,7 +139,8 @@ public class BallerinaHTTPConnectorListener implements HttpConnectorListener {
 
     private boolean checkForInterceptorDataBinding(HttpCarbonMessage inboundMessage, int interceptorServiceIndex,
                                                    InterceptorResource interceptorResource) {
-        if (HttpDispatcher.shouldDiffer(interceptorResource) && inboundMessage.isAccessedInNonInterceptorService()) {
+        if (!inboundMessage.isLastHttpContentArrived() && HttpDispatcher.shouldDiffer(interceptorResource) &&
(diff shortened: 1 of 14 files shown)


native/src/main/java/io/ballerina/stdlib/http/api/BallerinaHTTPConnectorListener.java
```diff
@@ -139,7 +139,8 @@ private boolean executeInterceptorServices(List<HTTPInterceptorServicesRegistry>
 
     private boolean checkForInterceptorDataBinding(HttpCarbonMessage inboundMessage, int interceptorServiceIndex,
                                                    InterceptorResource interceptorResource) {
-        if (HttpDispatcher.shouldDiffer(interceptorResource) && inboundMessage.isAccessedInNonInterceptorService()) {
+        if (!inboundMessage.isLastHttpContentArrived() && HttpDispatcher.shouldDiffer(interceptorResource) &&
```
chamil321, line 142: Looks like the isLastHttpContentArrived is synchronised. Shall we look for any performance change doing 10 mins load test?

chamil321: I think it's not needed as the same thread executes this invocation

chamil321 (approved):

native/src/main/java/io/ballerina/stdlib/http/api/BallerinaHTTPConnectorListener.java
TharmiganK: Created a task for this : [`Perf test for payload binding in HTTP service #3079`](https://github.com/ballerina-platform/ballerina-standard-library/issues/3079)

PR #1198, opened 2022-08-19
Make graceful stop timeout configurable

chamil321:
## Examples

```ballerina
isolated http:Listener someEP = check new(9095);

service /startService on http:Listener(9090, gracefulStopTimeout = 5) {


    resource function get testGracefully() returns string|error? {
        check someEP.gracefulStop();
        return "someEP is gracefully stopped!";
    }

    resource function get testImmediately() returns string|error? {
        check someEP.immediateStop();
        return "someEP is immediately stopped!";
    }
}
```

## Checklist
- [x] Linked to an issue
- [ ] Updated the changelog
- [x] Added tests
- [ ] Updated the spec


ballerina/http_constants.bal
```diff
@@ -30,8 +30,11 @@ const string HTTP_SCHEME = "http://";
 # Represents https protocol scheme
 const string HTTPS_SCHEME = "https://";
 
-# Constant for the default listener endpoint timeout
-const decimal DEFAULT_LISTENER_TIMEOUT = 120; //2 mins
+# Constant for the default listener endpoint timeout in seconds
+const decimal DEFAULT_LISTENER_TIMEOUT = 60;
```
ayeshLK, line 34: Any reason to reduce the default listener timeout ?

ayeshLK:
Shall we update the changelog as well ?

ballerina/http_constants.bal
chamil321: Decision was taken due to the current fast paced environments. 120 seconds of delayed response is quite high to expect a failure in an environment where the throughput is 20K. However this is configurable. Hence, depends on the requirement user can change it.

chamil321:
> Shall we update the changelog as well ?

Done via [ac3c082](https://github.com/ballerina-platform/module-ballerina-http/pull/1198/commits/ac3c082a85f3d84a89346f5ae84f5317210f6321)

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/listener/ServerConnectorBootstrap.java
```diff
@@ -297,7 +294,9 @@ private boolean unBindInterface() throws InterruptedException {
             ChannelFuture future = getChannelFuture();
             if (future != null) {
                 //Close will stop accepting new connections.
-                future.channel().close().sync();
+                ChannelFuture closeFuture = future.channel().close();
+                Thread.sleep(gracefulStopTimeout);
+                closeFuture.sync();
```
shafreenAnfar, line 299: I thought the code should be

Thread.sleep(gracefulStopTimeout);
future.channel().close().sync();

chamil321: If the sleep was executed before the close, unbind does not happen until the close is called. So it's just a thread block.

This grace period is intend to count after executing close() to provide some time to cater already accepted request. 
closeFuture.sync() is called afterward to wait till the close is succeeded.

shafreenAnfar (approved):

PR #1210, opened 2022-08-26
Fix spec inconsistencies in imperative client security

ayeshLK:
## Examples
N/A

## Checklist
- [ ] ~Linked to an issue~
- [ ] ~Updated the changelog~
- [ ] ~Added tests~
- [X] Updated the spec

### DIFF
--- a/docs/spec/spec.md
+++ b/docs/spec/spec.md
@@ -2487,7 +2487,8 @@ http:Client c = check new ("https://localhost:9090",
 );
 
 public function main() returns error? {
-    json response = check c->get("/foo/bar");
+    http:Request req = new;
+    json response = check c->post("/foo/bar", req);
     // evaluate response
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/spec/spec.md
```diff
@@ -2487,7 +2487,8 @@ http:Client c = check new ("https://localhost:9090",
 );
 
 public function main() returns error? {
-    json response = check c->get("/foo/bar");
+    http:Request req = new;
+    json response = check c->post("/foo/bar", req);
```
chamil321, line 2491: Shall we get rid of request and add simple json payload inline?

chamil321 (approved):

docs/spec/spec.md
ayeshLK: Since, we hv used `http:Request` in all security code sample we will use the same approach for consistency.

PR #1251, opened 2022-10-04
Fix CompilerPluginTest failure with lang 2201.3.0-20221004-103700-89b78c16

chamil321:
## Purpose
$subject

## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec


niveathika (commented):
Since the implementation changed based on lang version bump, http version should also be bumped to next version along with distribution in Ballerina.toml

niveathika (approved):

PR #1281, opened 2022-10-21
Fix dispatching logic issue due to double slash in basePath

chamil321:
## Examples
```ballerina
import ballerina/http;
import ballerina/test;

final http:Client clientEPTest = check new ("http://localhost:8080/");

service / on new http:Listener(8080) {

    resource function get foo() returns http:Ok {
        return {body:{name:"foo"}};
    }
}

@test:Config{}
public function testHttpEnumMethods() returns error? {
    json response = check clientEPTest->get("/foo");
    test:assertEquals(response, {name:"foo"});
}

```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] Updated the spec

### DIFF
--- a/native/src/main/java/io/ballerina/stdlib/http/api/HttpDispatcher.java
+++ b/native/src/main/java/io/ballerina/stdlib/http/api/HttpDispatcher.java
@@ -52,6 +52,8 @@ import java.util.Objects;
 import static io.ballerina.runtime.api.TypeTags.ARRAY_TAG;
 import static io.ballerina.stdlib.http.api.HttpConstants.DEFAULT_HOST;
 import static io.ballerina.stdlib.http.api.HttpConstants.EXTRA_PATH_INDEX;
+import static io.ballerina.stdlib.http.api.HttpConstants.HTTP_SCHEME;
+import static io.ballerina.stdlib.http.api.HttpConstants.SCHEME_SEPARATOR;
 import static io.ballerina.stdlib.http.api.HttpUtil.getParameterTypes;
 import static io.ballerina.stdlib.http.api.service.signature.ParamUtils.castParam;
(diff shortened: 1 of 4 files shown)


native/src/main/java/io/ballerina/stdlib/http/api/HttpDispatcher.java
```diff
@@ -159,15 +159,16 @@ public static InterceptorService findInterceptorService(HTTPInterceptorServicesR
...
         } catch (Exception e) {
             throw HttpUtil.createHttpError(e.getMessage(), HttpErrorType.SERVICE_DISPATCHING_ERROR);
         }
     }
 
-    private static void setInboundReqProperties(HttpCarbonMessage inboundReqMsg, URI requestUri, String basePath) {
-        String subPath = URIUtil.getSubPath(requestUri.getRawPath(), basePath);
+    private static void setInboundReqProperties(HttpCarbonMessage inboundReqMsg, URI requestUri,
+                                                String rawUri, String basePath) {
+        String subPath = URIUtil.getSubPath(rawUri, basePath);
```
TharmiganK, line 171: Passing the raw uri as the sub path is affecting the test cases where we have encoded urls and special characters.

Since the issue is with URI library which treats the double slash as the scheme separator. Can't we just pass a default scheme with the raw uri when creating the URI object? i.e. the following change is suggested :
```java
URI validatedUri = getValidatedURI("http://" + uriWithoutMatrixParams);
```
Since we are not using the scheme of this validatedUri, the value we set here does not matter. If we could get the scheme from the inbound request message, we can use that also.

chamil321: Thank tharmi, Yes, the issue is with the scheme separator. This is a much better solution

TharmiganK (approved):

PR #1383, opened 2022-12-21
Type reference support

chamil321:
## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec


gradle.properties
```diff
@@ -1,7 +1,7 @@
 org.gradle.caching=true
 group=io.ballerina.stdlib
-version=2.5.0-SNAPSHOT
-ballerinaLangVersion=2201.3.0-rc2.2
+version=2.6.0-SNAPSHOT
+ballerinaLangVersion=2201.4.0-20221116-082400-3eca35ee
```
HindujaB, line 4:
```suggestion
ballerinaLangVersion=2201.4.0-20221208-193400-e9fb14ee
```

niveathika: @HindujaB Shouldnt the stdlib versions be updated. AFAIK these timestamps are from type-reference branch and not the master

HindujaB: As per the offline discussion, we can go ahead with this time stamp, as the lang master is currently broken due to some other tests and the fix couldn't be merged without syncing std libraries. 
We will update the timestamp once we able to get master timestamp

warunalakshitha (approved):

PR #1417, opened 2023-01-25
Fix union types getting restricted by compiler plugin

dilanSachi:
## Purpose
$subject
Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/3929
## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] Updated the spec

### DIFF
--- a/ballerina-tests/http-service-tests/tests/anydata_union_return_type_service.bal
+++ b/ballerina-tests/http-service-tests/tests/anydata_union_return_type_service.bal
@@ -21,7 +21,7 @@ import http_service_tests.records;
 public type TestRecord records:RecordA|records:RecordA;
 
 service /unionReturnType on generalListener {
-    resource function unionResponse [string... path]() returns TestRecord[] {
+    resource function get unionResponse() returns TestRecord[] {
         records:RecordA response = {
             capacity: 10,
             elevationgain: 120,
--- a/ballerina-tests/http-service-tests/tests/anydata_union_return_type_service.bal
+++ b/ballerina-tests/http-service-tests/tests/anydata_union_return_type_service.bal
@@ -1,0 +1,49 @@
+// Copyright (c) 2023 WSO2 LLC. (http://www.wso2.com) All Rights Reserved.
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/http;
+import ballerina/test;
+import http_service_tests.records;
+
+public type TestRecord records:RecordA|records:RecordA;
+
+service /unionReturnType on generalListener {
+    resource function unionResponse [string... path]() returns TestRecord[] {
+        records:RecordA response = {
+            capacity: 10,
+            elevationgain: 120,
+            id: "2",
+            name: "Test",
+            night: false,
+            status: records:HOLD
+        };
+        return [response];
+    }
+}
+
+@test:Config {}
+function testUnionReturnType() returns error? {
+    http:Client 'client = check new("http://localhost:9000");
+    TestRecord[] response = check 'client->get("/unionReturnType/unionResponse");
+    test:assertEquals(response, [{
+        capacity: 10,
+        elevationgain: 120,
+        id: "2",
+        name: "Test",
+        night: false,
+        status: records:HOLD
+    }]);
+}
--- a/compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
+++ b/compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
@@ -149,7 +149,7 @@ public class HttpCompilerPluginUtil {
         } else if (kind == TypeDescKind.ARRAY) {
             TypeSymbol memberTypeDescriptor = ((ArrayTypeSymbol) returnTypeSymbol).memberTypeDescriptor();
             validateArrayElementTypeInReturnType(
-                    ctx, node, returnTypeStringValue, memberTypeDescriptor, diagnosticCode, isInterceptorType);
+                    ctx, node, returnTypeStringValue, memberTypeDescriptor, diagnosticCode);
         } else if (kind == TypeDescKind.TYPE_REFERENCE) {
             TypeSymbol typeDescriptor = ((TypeReferenceTypeSymbol) returnTypeSymbol).typeDescriptor();
             TypeDescKind typeDescKind = retrieveEffectiveTypeDesc(typeDescriptor);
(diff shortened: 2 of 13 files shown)


chamil321 (approved):

TharmiganK (commented):
Can we also add some ballerina tests to make sure the functionality also works?

dilanSachi:
> Can we also add some ballerina tests to make sure the functionality also works?

added a test case

ballerina-tests/http-service-tests/tests/anydata_union_return_type_service.bal
```diff
@@ -0,0 +1,49 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/http;
+import ballerina/test;
+import http_service_tests.records;
```
TharmiganK, line 19: Any specific reason to add the records to a separate module?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
```diff
@@ -149,7 +149,7 @@ public static void validateReturnType(SyntaxNodeAnalysisContext ctx, Node node,
         } else if (kind == TypeDescKind.ARRAY) {
             TypeSymbol memberTypeDescriptor = ((ArrayTypeSymbol) returnTypeSymbol).memberTypeDescriptor();
             validateArrayElementTypeInReturnType(
-                    ctx, node, returnTypeStringValue, memberTypeDescriptor, diagnosticCode);
+                    ctx, node, returnTypeStringValue, memberTypeDescriptor, diagnosticCode, isInterceptorType);
```
TharmiganK, line 152: Passing `isInterceptorType` along with the `validateArrayElementTypeInReturnType`, will allow developer to pass an array of `http:NextService` or an array of type which has a member `http:NextService`. So, lets remove that.

Seems like we are passing it in other places as well, like when the type is a map (Line #165). Can you change these as well?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
```diff
@@ -184,25 +184,31 @@ private static boolean isServiceType(TypeSymbol returnTypeSymbol) {
...
+            validateArrayElementTypeInReturnType(ctx, node, typeStringValue, typeSymbol,
+                    diagnosticCode, isInterceptorType);
         } else if (kind == TypeDescKind.TYPE_REFERENCE) {
             TypeSymbol typeDescriptor = ((TypeReferenceTypeSymbol) memberTypeDescriptor).typeDescriptor();
             TypeDescKind typeDescKind = retrieveEffectiveTypeDesc(typeDescriptor);
-            if (typeDescKind != TypeDescKind.RECORD && typeDescKind != TypeDescKind.MAP &&
-                    typeDescKind != TypeDescKind.TABLE) {
-                reportInvalidReturnType(ctx, node, typeStringValue, diagnosticCode);
+            if (typeDescKind == TypeDescKind.OBJECT) {
+                if (!isHttpModuleType(RESPONSE_OBJ_NAME, typeDescriptor)) {
```
TharmiganK, line 202: We don't allow array of `http:Response`. IIRC incase of `Object` type, we only allow `http:Response`. So in this case we can simply call `reportInvalidReturnType` for `Object` types.
@chamil321 can you also verify?

ballerina-tests/http-service-tests/tests/anydata_union_return_type_service.bal
dilanSachi: Added that to simulate the original use case

TharmiganK: Ah ok, in that case can we also run the `GraalVM Check` on your branch to see whether this is working without any issue?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
chamil321: Yes, we only return the `http:Response` as an object from the resource.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
dilanSachi: Updated

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
dilanSachi: Updated

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
dilanSachi: Passing https://github.com/dilanSachi/module-ballerina-http/actions/runs/4013386586

ballerina-tests/http-service-tests/tests/anydata_union_return_type_service.bal
dilanSachi: Passing https://github.com/dilanSachi/module-ballerina-http/actions/runs/4013386586

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
```diff
@@ -192,22 +192,58 @@ private static void validateArrayElementTypeInReturnType(SyntaxNodeAnalysisConte
...
+                    if (effectiveTypeDesc == TypeDescKind.OBJECT || effectiveTypeDesc == TypeDescKind.ERROR) {
+                        reportInvalidReturnType(ctx, node, typeStringValue, diagnosticCode);
+                        return;
+                    } else if (effectiveTypeDesc == TypeDescKind.TYPE_REFERENCE) {
+                        if (isStatusCodeResponseType((TypeReferenceTypeSymbol) typeSymbol)) {
+                            reportInvalidReturnType(ctx, node, typeStringValue, diagnosticCode);
+                            return;
+                        }
+                        Optional<String> definitionName = ((TypeReferenceTypeSymbol) typeSymbol).definition().getName();
+                        if (definitionName.isPresent() && definitionName.get().equals("StatusCodeResponse")) {
```
TharmiganK, line 219: What is the purpose of this check? Does this checks the `StatusCodeResponse` included records? i.e.
```bal
type AlbumConflict record {|
    *http:Conflict;
    record {
        string message;
    } body;
|};
```

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
```diff
@@ -192,22 +192,58 @@ private static void validateArrayElementTypeInReturnType(SyntaxNodeAnalysisConte
...
+            validateArrayElementTypeInReturnType(ctx, node, typeStringValue, memberTypeDescriptor,
+                    diagnosticCode);
         } else {
             reportInvalidReturnType(ctx, node, typeStringValue, diagnosticCode);
         }
     }
 
+    public static boolean isStatusCodeResponseType(TypeReferenceTypeSymbol typeSymbol) {
+        if (typeSymbol.getName().isPresent()) {
+            return typeSymbol.getName().get().equals("StatusCodeResponse");
```
TharmiganK, line 242: Can we make this `StatusCodeResponse` string as a constant and reuse?

TharmiganK, line 242: Also we should check whether this `StatusCodeResponse` record is from `http` package. So can't we reuse the `isHttpModuleType` function here?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
dilanSachi: It only checks for `StatusCodeResponse` type

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
dilanSachi: Updated with the suggestion

dilanSachi: Updated with the suggestion

TharmiganK (approved):

PR #1428, opened 2023-02-05
Add code modifier for @http:payload annotation injection

chamil321:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/3276

## Examples

When the following conditions are met, the listener identifies the default payload parameter, which is defined 
without the @http:Payload annotation:
- The default payload parameter rules are only applicable to POST, PUT, PATCH, DELETE, and DEFAULT accessors.
- Parameters must contain only one structured(map/record/table/tuple/array) type or xml. However, byte[] is an exception,
  and it is considered as a payload param.
    - resource function post(Student[] p) {} -> Student[] is payload param type
    - resource function post(int[] p) {} -> int[] is query param type
    - resource function post(int p) {} -> int is query param type
    - resource function post(Student p) {} -> Student is payload param type
- If there's more than one structured type, the ambiguity must be resolved using the @http:Payload annotation.
    - resource function post(@http:Payload Student p, map<json> q) {} -> p is payload, q is query
- If there are no structured types, all parameters are considered query parameters.
    - resource function post(@http:Payload string p, string q) {}
- If the query parameter is structured, then the @http:Query annotation is required.
    - resource function post(@http:Query map<json> p) {}
- The only types allowed in the union for a parameter are structured types, xml, and nil.
    - resource function post(Student|xml p) {} -> Student|xml is payload param type
    - resource function post(map<json>|xml p) {} -> map<json>|xml is payload param type
    - resource function post(Student? p) {} -> Student? is payload param type
    
```ballerina
import ballerina/http;

type Person record {|
    readonly int id;
|};

service on ep {

    resource function post person(string? xyz, Person abc) returns Person {
       return abc;
    }
}
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec

### DIFF
--- a/compiler-plugin-tests/src/test/resources/ballerina_sources/sample_package_31/service.bal
+++ b/compiler-plugin-tests/src/test/resources/ballerina_sources/sample_package_31/service.bal
@@ -1,8 +1,4 @@
-<<<<<<< HEAD
-// Copyright (c) 2023 WSO2 LLC. (http://www.wso2.com) All Rights Reserved.
-=======
 // Copyright (c) 2023, WSO2 LLC. (http://www.wso2.org) All Rights Reserved.
->>>>>>> Add test cases
 //
 // WSO2 LLC. licenses this file to you under the Apache License,
 // Version 2.0 (the "License"); you may not use this file except
--- a/compiler-plugin-tests/src/test/resources/ballerina_sources/sample_package_31/service.bal
+++ b/compiler-plugin-tests/src/test/resources/ballerina_sources/sample_package_31/service.bal
@@ -1,4 +1,8 @@
+<<<<<<< HEAD
 // Copyright (c) 2023 WSO2 LLC. (http://www.wso2.com) All Rights Reserved.
+=======
+// Copyright (c) 2023, WSO2 LLC. (http://www.wso2.org) All Rights Reserved.
+>>>>>>> Add test cases
 //
 // WSO2 LLC. licenses this file to you under the Apache License,
 // Version 2.0 (the "License"); you may not use this file except
--- a/compiler-plugin-tests/src/test/resources/ballerina_sources/sample_package_31/service.bal
+++ b/compiler-plugin-tests/src/test/resources/ballerina_sources/sample_package_31/service.bal
@@ -1,4 +1,4 @@
-// Copyright (c) 2023, WSO2 LLC. (http://www.wso2.org) All Rights Reserved.
+// Copyright (c) 2023 WSO2 LLC. (http://www.wso2.com) All Rights Reserved.
 //
 // WSO2 LLC. licenses this file to you under the Apache License,
 // Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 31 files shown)


compiler-plugin-tests/src/test/resources/ballerina_sources/sample_package_31/service.bal
```diff
@@ -0,0 +1,51 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/http;
+import ballerina/mime;
```
TharmiganK, line 18:
```suggestion
```
Unused module `mime`. Is this intentional?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/codemodifier/HttpPayloadParamIdentifier.java
```diff
@@ -0,0 +1,271 @@
...
+                return;
+            }
+        }
+
+        for (ParamData nonAnnotatedParam : nonAnnotatedParams) {
+            ParameterSymbol parameterSymbol = nonAnnotatedParam.getParameterSymbol();
+            if (validateNonAnnotatedParams(ctx, parameterSymbol.typeDescriptor(),
+                                           paramAvailability)) { //TODO change name of the func
+                paramAvailability.setPayloadParamSymbol(parameterSymbol);
+                this.payloadParamContextMap.put(ctx.documentId(),
```
TharmiganK, line 148: Using the same key will overwrite the values, so if we have multiple resources in a service this will effectively change only one resource parameter

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/codemodifier/HttpPayloadParamIdentifier.java
```diff
@@ -0,0 +1,271 @@
...
+import static io.ballerina.stdlib.http.compiler.HttpCompilerPluginUtil.updateDiagnostic;
+
+
+/**
+ * {@code HttpPayloadParamIdentifier} identifies the payload param during the initial syntax node analysis.
+ *
+ * @since 2201.5.0
+ */
+public class HttpPayloadParamIdentifier extends HttpServiceValidator {
+    private final Map<DocumentId, PayloadParamContext> payloadParamContextMap;
```
TharmiganK, line 67: Since this `payloadParamContextMap` is created for each service declarations and the documentId is same for one service. But we can have many resources which may need `payload` annotation.

So I think we should change this `Map<DocumentId, List<PayloadParamContext>>`. WDTY?

chamil321: Yes, I refactored the classes to be more descriptive and accurate. Please review again

shafreenAnfar (approved):
APIs looks good to me.

TharmiganK (approved):
LGTM

PR #1455, opened 2023-02-22
Add `:authority` header to push promise headers

dilanSachi:
## Purpose
$subject
Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/3077
## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] Updated the spec

### DIFF
--- a/native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/common/states/Http2StateUtil.java
+++ b/native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/common/states/Http2StateUtil.java
@@ -217,7 +217,8 @@ public class Http2StateUtil {
         HttpRequest httpRequest = pushPromise.getHttpRequest();
         httpRequest.headers().add(HttpConversionUtil.ExtensionHeaderNames.SCHEME.text(), HTTP_SCHEME);
         if (!httpRequest.headers().contains(HttpHeaderNames.HOST)) {
-            String host = ((InetSocketAddress) inboundRequestMsg.getProperties().get(LOCAL_ADDRESS)).getHostName();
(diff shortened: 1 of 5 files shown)


native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/common/states/Http2StateUtil.java
```diff
@@ -215,9 +216,12 @@ public static void writeHttp2Promise(Http2PushPromise pushPromise, ChannelHandle
         // Construct http request
         HttpRequest httpRequest = pushPromise.getHttpRequest();
         httpRequest.headers().add(HttpConversionUtil.ExtensionHeaderNames.SCHEME.text(), HTTP_SCHEME);
+        if (!httpRequest.headers().contains(HttpHeaderNames.HOST)) {
+            String host = ((InetSocketAddress) inboundRequestMsg.getProperties().get(LOCAL_ADDRESS)).getHostName();
```
chamil321, line 220: Calling the InetSocketAddress.getHostName() method can potentially incur a cost due to a reverse DNS lookup.
Can we use the InetSocketAddress.getAddress() method instead, which returns the IP address? would it be enough to satisfy the client?

dilanSachi: Updated to use `getAddress` https://github.com/ballerina-platform/module-ballerina-http/pull/1455/commits/11ee451e896de4c40b6bb0d750a44cdf954c7e32

chamil321 (approved):

PR #1460, opened 2023-02-23
Fix the `ReqCtxMember` type definition

TharmiganK:
## Purpose

> $Subject

Related to: https://github.com/ballerina-platform/ballerina-lang/pull/39573

## Examples

N/A

## Checklist
- [ ] <s>Linked to an issue</s>
- [ ] <s>Added tests</s>
- [ ] <s>Updated the spec</s>
- [ ] <s>Checked native-image compatibility</s>

### DIFF
--- a/ballerina/http_request_context.bal
+++ b/ballerina/http_request_context.bal
@@ -89,7 +89,7 @@ public isolated class RequestContext {
     # + return - Attribute value or an error. The error is returned if the member does not exist or
     #  if the member value is not of the expected type
     public isolated function getWithType(string key, ReqCtxMemberType targetType = <>)
-        returns targetType|error = @java:Method {
+        returns targetType|ListenerError = @java:Method {
         'class: "io.ballerina.stdlib.http.api.nativeimpl.ExternRequestContext"
     } external;
 
--- a/ballerina/http_request_context.bal
+++ b/ballerina/http_request_context.bal
@@ -15,10 +15,13 @@
 // under the License.
 
 import ballerina/jballerina.java;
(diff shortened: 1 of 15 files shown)


ballerina/http_request_context.bal
```diff
@@ -15,10 +15,13 @@
 // under the License.
 
 import ballerina/jballerina.java;
-import ballerina/lang.value;
+
+// This is same as the `value:Cloneable`, except that it does not include `error` type.
+# Represents a non-error type that can be cloned.
+public type Cloneable (any & readonly)|xml|Cloneable[]|map<Cloneable>|table<map<Cloneable>>;
```
MaryamZi, line 21: This will also disallow `error[]`, `map<error>`, etc, although they are disjoint with the error type. Hope that's OK.

MaryamZi: If we want to allow them, we can probably use 

```suggestion
public type Cloneable (any & readonly)|xml|value:Cloneable[]|map<value:Cloneable>|table<map<value:Cloneable>>;
```

TharmiganK: Most of the time `RequestContext` is used to store the `any data` type. But we kept it to support all the general types. IMO we don't need to support structure types of `error`. (But anyway I will update the spec with this information)
@chamil321 @shafreenAnfar WDYT?

shafreenAnfar: Yeah, I also think that passing around error is not useful, therefore, lets revisit this if there is some requirement .

chamil321 (approved):

PR #1470, opened 2023-02-28
Add JWT information into `http:RequestContext`

dilanSachi:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/3408
## Examples
```bal
service class RequestInterceptorJwtInformation {
    *http:RequestInterceptor;

    resource function 'default [string... path](http:RequestContext ctx, http:Request req) returns http:NextService|error? {
        [jwt:Header, jwt:Payload]|error? result = ctx.getJWTInfo();
        if result is [jwt:Header, jwt:Payload] {
            
        } else if result is error {
            // handle the error
        }
        return ctx.next();
    }
}
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec


ballerina/http_interceptors.bal
```diff
@@ -51,3 +53,19 @@ service class DefaultErrorInterceptor {
...
+
+    # Gets invoked during object initialization.
+    #
+    isolated function init() {}
+
+    # Decodes a jwt string to `[jwt:Header, jwt:Payload]`.
+    #
+    # + jwt - The jwt value as a string
+    # + return - `[jwt:Header, jwt:Payload]` if successful or else an `error`
+    isolated function decodeJwt(string jwt) returns [jwt:Header, jwt:Payload]|error {
```
TharmiganK, line 68: Do we define this function only to be used by the native invoke call?

ballerina-tests/http-interceptor-tests/tests/test_common.bal
```diff
@@ -373,6 +373,16 @@ service class RequestInterceptorUserAgentField {
     }
 }
 
+service class RequestInterceptorJwtInformation {
+    *http:RequestInterceptor;
+
+    resource function 'default [string... path](http:RequestContext ctx, http:Request req) returns http:NextService|error? {
+        reqCtxJwtValues[0] = check ctx.getWithType(http:JWT_HEADER);
```
TharmiganK, line 380: Shouldn't we introduce new function(s) on Request Context object to retrieve these information(this will forbid users to override these internal values)  rather than using the members? We can have a function like `getJWTInfo()` and return `[jwt:Header, jwt:Payload]|error`.
@chamil321 @shafreenAnfar WDYT?

native/src/main/java/io/ballerina/stdlib/http/api/HttpDispatcher.java
```diff
@@ -424,6 +444,52 @@ static BObject createRequestContext(HttpCarbonMessage httpCarbonMessage) {
...
+            BMap requestCtxMembers = requestContext.getMapValue(REQUEST_CTX_MEMBERS);
+            requestCtxMembers.put(JWT_HEADER, ((BArray) decodedJwt).get(0));
+            requestCtxMembers.put(JWT_PAYLOAD, ((BArray) decodedJwt).get(1));
+        }
+    }
+
+    private static Object invokeJwtDecode(Runtime runtime, String authHeader) {
+        final Object[] jwtValues = new Object[1];
+        CountDownLatch countDownLatch = new CountDownLatch(1);
+        Callback decodeCallback = new Callback() {
```
TharmiganK, line 459: Are we silently skipping the error(JWT decode Error) returned from the invoke call? 
Also is it possible to add a negative test case where an error is returned from this async call?

dilanSachi: Yes. Since there is a posibility of `Bearer ....` header to be both `jwt` or `oauth2`, thought of ignoring the error.

ballerina/http_interceptors.bal
dilanSachi: Yes. AFAIK inorder to call a ballerina method by the runtime, the method needs to reside inside a class.
https://github.com/ballerina-platform/ballerina-lang/issues/39743

TharmiganK: The concern is if since this is `public` class this is exposed to the user, but there is no actual use case in the user's point-of-view. Also not sure whether adding and additional async call will cause any unexpected errors. @chamil321 WDYT?

Can't we call the JWT native methods in our native code to do `jwt:decode`? This will add the jwt native jar as a dependency of http native. @ayeshLK is there a possibility to this?

dilanSachi: `jwt:decode` has a ballerina implementation.

dilanSachi: We can make the class private. Adding `public` is a mistake.

dilanSachi: Regarding the `async` call, since `jwt:decode` is an isolated function, it won't have any issues right?

ballerina-tests/http-interceptor-tests/tests/test_common.bal
TharmiganK: If we use a separate function, then we can set the jwt string as an attribute of the request context object and do the decoding in the ballerina side :
```bal
public isolated class RequestContext {
    private final string? JWTString = (); // Need to set this value from the native side

    ......

    function getJWTInfo() returns [jwt:Header,  jwt:Payload]|error? {
        if self.JWTString is string {
            // Have to return a HttpError by adding the jwt:Error as cause
            return jwt:decode(<string> self.JWTString);
        }
        return ();
    }
}
```

dilanSachi: Updated as discussed

chamil321 (approved):

ballerina/http_request_context.bal
```diff
@@ -105,10 +105,30 @@ public isolated class RequestContext {
...
+    # Provides the JWT information from the request.
+    #
+    # + return - `[jwt:Header, jwt:Payload]` if decoding the header is successful, `error` if any error occurs
+    #             while decoding, or `nil` if no jwt header found.
+    public function getJWTInfo() returns [jwt:Header, jwt:Payload]|error? {
+        string? authString = check self.getAuthString();
+        if authString is string {
+            [jwt:Header, jwt:Payload]|jwt:Error result = jwt:decode(<string> authString);
+            if result is jwt:Error {
+                return error ListenerAuthError("An error occured while decoding jwt string.", result);
```
TharmiganK, line 117: We usually start the error message in the lowercase
```suggestion
                return error ListenerAuthError("an error occured while decoding jwt string.", result);
```

docs/spec/spec.md
```diff
@@ -2138,15 +2138,23 @@ public isolated class RequestContext {
...
+    # + return - `[jwt:Header, jwt:Payload]` if decoding the header is successful, `error` if any error occurs
+    #             while decoding, or `nil` if no jwt header found.
+    public function getJWTInfo() returns [jwt:Header, jwt:Payload]|error? {}
 
     # Calls the next service in the interceptor pipeline.
     #
     # + return - The next service object in the pipeline. An error is returned, if the call fails
     public isolated function next() returns NextService|error? = external;
+
+    isolated function getAuthString() returns string|error? = @java:Method {}
```
TharmiganK, line 2157: This we don't need to expose through spec since the function is not public.

native/src/main/java/io/ballerina/stdlib/http/api/HttpDispatcher.java
```diff
@@ -414,6 +415,11 @@ static BObject getCaller(Resource resource, HttpCarbonMessage httpCarbonMessage,
 
     static BObject createRequestContext(HttpCarbonMessage httpCarbonMessage) {
         BObject requestContext = ValueCreatorUtils.createRequestContextObject();
+        String authHeader = httpCarbonMessage.getHeader(AUTHORIZATION_HEADER);
+        if (Objects.nonNull(authHeader) && authHeader.contains("Bearer ")) {
```
TharmiganK, line 419: Can we use a constant for `"Bearer "`?

native/src/main/java/io/ballerina/stdlib/http/api/HttpDispatcher.java
```diff
@@ -414,6 +415,11 @@ static BObject getCaller(Resource resource, HttpCarbonMessage httpCarbonMessage,
...
+            requestContext.addNativeData(HttpConstants.AUTHORIZATION_STRING,
+                    StringUtils.fromString(authHeader.split(" ")[1]));
```
TharmiganK, line 421: Can we use a constant for this as well?

native/src/main/java/io/ballerina/stdlib/http/api/ValueCreatorUtils.java
```diff
@@ -109,19 +109,17 @@ public static BMap<BString, Object> createHTTPRecordValue(String recordTypeName)
...
-    private static BObject createObjectValue(Module module, String objectTypeName,
-                                             Object... fieldValues) {
-
-        Object[] fields = new Object[fieldValues.length * 2];
-
-        // Adding boolean values for each arg
-        for (int i = 0, j = 0; i < fieldValues.length; i++) {
-            fields[j++] = fieldValues[i];
-            fields[j++] = true;
+    private static BObject createObjectValue(Module module, String objectTypeName, Object... fieldValues) {
```
TharmiganK, line 112: Is this change needed? Even if we fix this from http, runtime does not support this right?

dilanSachi: There was an issue in how this was implemented. Earlier, we have passed `ValueCreator.createObjectValue(module, objectTypeName, null, null, null, fields);` mistakenly. But the `ValueCreator` API is `public static BObject createObjectValue(Module packageId, String objectTypeName, Object... fieldValues) {`

native/src/main/java/io/ballerina/stdlib/http/api/HttpDispatcher.java
dilanSachi: Updated in https://github.com/ballerina-platform/module-ballerina-http/pull/1470/commits/461c0635c2a83486b45f6fa21243bf23fbeb8227

native/src/main/java/io/ballerina/stdlib/http/api/HttpDispatcher.java
dilanSachi: Updated in https://github.com/ballerina-platform/module-ballerina-http/pull/1470/commits/461c0635c2a83486b45f6fa21243bf23fbeb8227

docs/spec/spec.md
dilanSachi: Updated in https://github.com/ballerina-platform/module-ballerina-http/pull/1470/commits/461c0635c2a83486b45f6fa21243bf23fbeb8227

TharmiganK (approved):
LGTM

PR #1490, opened 2023-03-14
Fix the URL path encoding issue

Bhashinee:
## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec


Bhashinee:
Introducing this library aligns with the URI specification. But there are users who already use encoded path segments. So introducing this library will break the existing APIs. Because of that need to merge this with care. Hence keeping as a closed PR.

PR #1552, opened 2023-04-11
HTTP interceptor related improvements

TharmiganK:
## Purpose

> $Subject

- Fixes: https://github.com/ballerina-platform/ballerina-standard-library/issues/3964
- Fixes: https://github.com/ballerina-platform/ballerina-standard-library/issues/3969

## Examples

1. Allows configuring interceptor pipeline with a single interceptor :

```bal
listener http:Listener serverEP = new (9090, interceptors = new Interceptor1());

@http:ServiceConfig {
    interceptors: new Interceptor2()
}
service on serverEP {

    resource function get test() returns string? {
        return "Hello, World!";
    }
}
```

2. Allows `http:Request` in the `ResponseInterceptor` and the `ResponseErrorInterceptor` :

```bal
service class ResponseInterceptorWithReq {
    *http:ResponseInterceptor;

    remote function interceptResponse(http:Request req, http:Response res, http:RequestContext ctx)
             returns http:NextService|error? {
         ...
    }
}

service class ResponseErrorInterceptorWithReq {
    *http:ResponseErrorInterceptor;

    remote function interceptResponseError(error err, http:Request req)
             returns http:NextService|error? {
         ...
    }
}
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec
- [x] [Checked native-image compatibility](https://github.com/ballerina-platform/module-ballerina-http/actions/runs/4717562459)


native/src/main/java/io/ballerina/stdlib/http/api/HttpResponseInterceptorUnitCallback.java
```diff
@@ -63,6 +66,9 @@ public void notifySuccess(Object result) {
             invokeErrorInterceptors((BError) result, true);
             return;
         }
+        if (possibleLastInterceptor) {
```
shafreenAnfar, line 69: What is not `lastInterceptor` but `possibleLastInterceptor `?

TharmiganK: Since we are not aware of the logic inside the last response interceptor in the pipeline, it is not possible to determine the last interceptor (It is possible to return an error from the last response interceptor, then the last interceptor should be a response error interceptor). That's why I mark the interceptors as `possibleLastInterceptors` and clean the request message only when the return type is not `error` otherwise the request content will be released in the next response error interceptors.

shafreenAnfar (approved):

PR #1579, opened 2023-04-19
Add a load test with interceptors

TharmiganK:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/4343

## Examples

N/A

## Checklist
- [x] Linked to an issue
- [ ] ~Updated the changelog~
- [ ] ~Added tests~
- [ ] ~Updated the spec~


dilanSachi:
Did we find the root cause for connection failures when we directly connect to the backend?

TharmiganK:
> Did we find the root cause for connection failures when we directly connect to the backend?

The failures may be due to the pod memory limitation we still have around 2%. I didn't check by directly making request to the backend. I used a passthrough since we are making requests with different scenarios to test different interceptor execution paths. I tried to use multiple requests using the jmeter but it seems there is an issue with the workflow, the workflow only adds the results of one request.

load-tests/interceptors_passthrough/src/utils.bal
```diff
@@ -0,0 +1,69 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/http;
+
+const string CLIENT_URL = "http://interceptors:9090";
+final http:Client testClient = check new (CLIENT_URL);
```
dilanSachi, line 20: Even if we use 60 concurrent users, we will be using the same client right?
hope thats ok

PR #1596, opened 2023-04-26
Add max header size validation for http2

dilanSachi:
## Purpose
$subject
Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/489
## Checklist
- [x] Linked to an issue
- [ ] Updated the changelog
- [x] Added tests
- [ ] Updated the spec


dilanSachi:
Had to update the test cases to increase the allowed header size limits as per this comment.
https://github.com/ballerina-platform/ballerina-standard-library/issues/489#issuecomment-1523035486

ballerina/http_service_endpoint.bal
```diff
@@ -203,7 +203,8 @@ public type ListenerHttp1Settings record {|
 # Provides inbound request URI, total header and entity body size threshold configurations.
 #
 # + maxUriLength - Maximum allowed length for a URI. Exceeding this limit will result in a `414 - URI Too Long`
-#                  response
+#                  response. For HTTP/2, this limit will not be applicable as it already has a `:path`
```
TharmiganK, line 206: Isn't it better to make the existing one as a part of `ListenerHttp1Settings`?
```bal
public type Http1RequestLimitConfigs record {|
    int maxUriLength = 4096;
    int maxHeaderSize = 8192;
    int maxEntityBodySize = -1;
|};

public type ListenerHttp1Settings record {|
    KeepAlive keepAlive = KEEPALIVE_AUTO;
    int maxPipelinedRequests = MAX_PIPELINED_REQUESTS;
    Http1RequestLimitConfigs requestLimits;
|};
```

Then we can deprecate this `maxUriLength` field here.

dilanSachi: We can do that. However it will be confusing for the user when switching between http and http/2 right?
Defining `maxHeaderSize` for http/2 in default `http:ListenerConfiguration` and defining `maxHeaderSize` for http1 in `ListenerHttp1Settings`....

dilanSachi: Decided to keep the configs as it is because if we move the urilength request limit outside, the config `RequestLimitConfigs` will not be meaningful anymore.

TharmiganK (approved):
LGTM

PR #1603, opened 2023-05-01
Rewrite the logic for populating resource parameter

TharmiganK:
## Purpose

> $Subject

Fixes : https://github.com/ballerina-platform/ballerina-standard-library/issues/4372
Fixes : https://github.com/ballerina-platform/ballerina-standard-library/issues/4374

The following issues were found in the previous resource parameter(path, query and header) binding logic:
- We had a logic of verifying the path/query/header parameter based on the type tag of the parameter type
   >Because of this we had some issues when the parameter type is enum or finite. And time-to-time we had updated this logic to support specific types when there is a bug reported. Ideally, we should have verified this logic rather than adding support for specific cases.

- We parsed the path/query/header parameter based on the parameter type.
   >Because of this when we support new types, we ended up changing this parsing logic as well.
  
- After parsing the query parameter to one of this type: `string`, `int`, `float`, `decimal`, `boolean`, `map<json>`, we have returned this raw type to the resource function rather than converting to the type specified in the resource signature.
   >Because of this the type information is loss inside the resource function. And when we have readonly types we have to specifically made these types readonly from our native.
   
Ideally, when we bind a path/query/header parameter we should have done the following:
1. Get the effective type from the parameter type. It can be one of the raw type: `string`, `int`, `float`, `decimal`, `boolean` and `map<json>`(only for query param).
2. In order to parse the path/query/header string we only need this effective type and not the parameter type.
3. After parsing the parameter to one of the raw type, we should convert the value to the parameter type using a runtime API like : `ValueUtils.convert` it handles all the readonly and union resolution and will throw a `BError` when the conversion fails.

So this PR is to rewrite the resource parameter binding logic as described above.

## Examples

With this rewrite, the union types are allowed in as path/query/header parameters only if the effective type is one of the allowed raw type.

```bal

enum EnumValue {
    VALUE1,
    VALUE2,
    VALUE3
}

type FiniteIntTypes 1|2|3|4;

service on new http:Listener(9090) {

    // This is allowed since the effective type is string
    resource function get case1/[EnumValue|"VALUE4"|"VALUE5" path]() {}
    
    // This is allowed since the effective type is integer
    resource function get case2(FiniteIntTypes|5|6|7 query) {}

    // This is not allowed
    resource function get case3/[EnumValue|1|2 path]() {}
    
    // Lets say Record1 and Record2 are effectively map of json types
    // Then this is allowed
    // If the records are structurally same, then it will bind to the first type matched in the union
    resource function get case4(Record1|Record2 query) {}
}
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] ~Updated the spec~

### DIFF
--- a/compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/CompilerPluginTest.java
+++ b/compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/CompilerPluginTest.java
@@ -774,4 +774,12 @@ public class CompilerPluginTest {
                 " 'string','int','float','decimal','boolean' types or a rest parameter with one of the above types",
                 HTTP_145);
     }
+
+    @Test
+    public void testResourceSignatureParamValidations() {
+        Package currentPackage = loadPackage("sample_package_38");
+        PackageCompilation compilation = currentPackage.getCompilation();
+        DiagnosticResult diagnosticResult = compilation.diagnosticResult();
+        Assert.assertEquals(diagnosticResult.errorCount(), 36);
+    }
 }
--- a/compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/CompilerPluginTest.java
+++ b/compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/CompilerPluginTest.java
@@ -781,5 +781,101 @@ public class CompilerPluginTest {
         PackageCompilation compilation = currentPackage.getCompilation();
         DiagnosticResult diagnosticResult = compilation.diagnosticResult();
         Assert.assertEquals(diagnosticResult.errorCount(), 36);
+        assertError(diagnosticResult, 0, "invalid union type of header param 'header': expected one of the" +
+                " 'string','int','float','decimal','boolean' types, an array of the above types or a record which " +
+                "consists of the above types can only be union with '()'. Eg: string|() or string[]|()", HTTP_110);
+        assertError(diagnosticResult, 1, "invalid type of header param 'header': expected one of the following" +
+                " types is expected: 'string','int','float','decimal','boolean', an array of the above types or a " +
+                "record which consists of the above types", HTTP_109);
+        assertError(diagnosticResult, 2, "invalid union type of header param 'header': expected one of the" +
+                " 'string','int','float','decimal','boolean' types, an array of the above types or a record which " +
+                "consists of the above types can only be union with '()'. Eg: string|() or string[]|()", HTTP_110);
+        assertError(diagnosticResult, 3, "invalid type of header param 'header': expected one of the following" +
+                " types is expected: 'string','int','float','decimal','boolean', an array of the above types or a " +
+                "record which consists of the above types", HTTP_109);
+        assertError(diagnosticResult, 4, "invalid type of header param 'header': expected one of the following" +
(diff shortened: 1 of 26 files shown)


shafreenAnfar (approved):
LGTM!

native/src/main/java/io/ballerina/stdlib/http/api/service/signature/ParamUtils.java
```diff
@@ -109,6 +134,173 @@ private static BArray getBArray(String[] valueArray, ArrayType arrayType, Type e
...
+                return getEffectiveTypeTagFromIntersectionType((IntersectionType) referredType, originalType,
+                        paramType);
+        }
+        throw HttpUtil.createHttpError("invalid " + paramType + " parameter type '" + originalType + "'");
+    }
+
+    private static int getEffectiveTypeTagFromArrayType(ArrayType type, Type originalType, String paramType) {
+        Type elementType = TypeUtils.getReferredType(type.getElementType());
+        int elementTypeTag = elementType.getTag();
+        if (elementTypeTag == ARRAY_TAG) {
```
dilanSachi, line 246:
```suggestion
        if (elementType.getTag() == ARRAY_TAG) {
```
redundant variable

compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/CompilerPluginTest.java
```diff
@@ -774,4 +774,12 @@ public void testPathParameterType() {
                 " 'string','int','float','decimal','boolean' types or a rest parameter with one of the above types",
                 HTTP_145);
     }
+
+    @Test
+    public void testResourceSignatureParamValidations() {
+        Package currentPackage = loadPackage("sample_package_38");
+        PackageCompilation compilation = currentPackage.getCompilation();
+        DiagnosticResult diagnosticResult = compilation.diagnosticResult();
+        Assert.assertEquals(diagnosticResult.errorCount(), 36);
```
dilanSachi, line 783: dont we need to validate the error messages?

TharmiganK: Will add the error message assertions 👍🏼

native/src/main/java/io/ballerina/stdlib/http/api/service/signature/ParamUtils.java
TharmiganK: Fixed via [5103338](https://github.com/ballerina-platform/module-ballerina-http/pull/1603/commits/5103338ef6638d45a4556aafe2ece8e5bf054d10)

compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/CompilerPluginTest.java
TharmiganK: Fixed via [bef7551](https://github.com/ballerina-platform/module-ballerina-http/pull/1603/commits/bef7551ee5309fd972cc7846e72f1ed3e3f56713)

dilanSachi (approved):

TharmiganK:
Skipping the GraalVM Check since it is already tested in the previous commit and the last commit is related to compiler plugin tests.

PR #1675, opened 2023-06-21
Update vulnerable netty version

ayeshLK:
## Examples
N/A

## Checklist
- [ ] ~Linked to an issue~
- [ ] ~Updated the changelog~
- [ ] ~Added tests~
- [ ] ~Updated the spec~


gradle.properties
```diff
@@ -4,7 +4,7 @@ version=2.9.0-SNAPSHOT
 ballerinaLangVersion=2201.7.0-20230619-175900-bb4e4544
 ballerinaTomlParserVersion=1.2.2
 commonsLang3Version=3.8.1
-nettyVersion=4.1.86.Final
+nettyVersion=4.1.94.Final
 nettyTcnativeVersion=2.0.54.Final
```
TharmiganK, line 8: We also need to update the `nettyTcnativeVersion` to the latest version which is compatible with this netty version - `2.0.61.Final`

PR #1706, opened 2023-07-12
Update implementation to resolve lang spec deviation misuse

dilanSachi:
## Purpose
Related issue - https://github.com/ballerina-platform/ballerina-standard-library/issues/4584
Also fixes - https://github.com/ballerina-platform/ballerina-standard-library/issues/4649
We have moved the interceptor array creation to a separate function instead of the `init` function and called it from native side when setting up the interceptors.

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] Updated the spec


TharmiganK (approved):
LGTM

PR #1718, opened 2023-07-20
Allow open records as query param

TharmiganK:
## Purpose

> $Subject

> Fixes: https://github.com/ballerina-platform/ballerina-standard-library/issues/4541

## Examples

```bal
import ballerina/http;

type User record {
    int id;
    string name;
};

type UserNew record {
    int id;
    string name;
    Address address;
};

type Address record {
    int no;
    string street;
};

service /sample on new http:Listener(9090) {
    
    resource function get path1(User user) {}

    resource function post path2(@http:Query User user) {}

    resource function put path3(@http:Query record {int id; string name;} user) {}

    resource function get path4(record{|User...;|} user) {}

    resource function post path5(@http:Query UserNew user) {}
}
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec

### DIFF
--- a/compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/Constants.java
+++ b/compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/Constants.java
@@ -60,7 +60,7 @@ public class Constants {
     public static final String NILABLE_FLOAT_ARRAY = "float[]?";
     public static final String NILABLE_DECIMAL_ARRAY = "decimal[]?";
     public static final String NILABLE_BOOLEAN_ARRAY = "boolean[]?";
-    public static final String NILABLE_MAP_OF_ANYDATA_ARRAY = "map<json>[]?";
+    public static final String NILABLE_MAP_OF_ANYDATA_ARRAY = "map<anydata>[]?";
 
     public static final String RESOURCE_RETURN_TYPE = "ResourceReturnType";
     public static final String INTERCEPTOR_RESOURCE_RETURN_TYPE = "InterceptorResourceReturnType";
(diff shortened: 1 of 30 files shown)


compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/Constants.java
```diff
@@ -55,13 +54,13 @@ public class Constants {
...
     public static final String NILABLE_BOOLEAN = "boolean?";
-    public static final String NILABLE_MAP_OF_JSON = "map<json>?";
+    public static final String NILABLE_MAP_OF_ANYDATA = "map<anydata>?";
     public static final String NILABLE_STRING_ARRAY = "string[]?";
     public static final String NILABLE_INT_ARRAY = "int[]?";
     public static final String NILABLE_FLOAT_ARRAY = "float[]?";
     public static final String NILABLE_DECIMAL_ARRAY = "decimal[]?";
     public static final String NILABLE_BOOLEAN_ARRAY = "boolean[]?";
-    public static final String NILABLE_MAP_OF_JSON_ARRAY = "map<json>[]?";
+    public static final String NILABLE_MAP_OF_ANYDATA_ARRAY = "map<json>[]?";
```
dilanSachi, line 63:
```suggestion
    public static final String NILABLE_MAP_OF_ANYDATA_ARRAY = "map<anydata>[]?";
```
should be `anydata` right?

TharmiganK: Yes, will fix that 👍🏼

dilanSachi (approved):
LGTM

PR #1729, opened 2023-07-26
Remove interceptor config from listener and service configurations

dilanSachi:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-standard-library/issues/4680

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] ~Updated the spec~


TharmiganK (approved):
LGTM

Since we are also removing the `interceptors` field from the `ListenerConfiguration` it can affect the connectors which have their own configuration by including `http:ListenerConfiguration`. Can we check on that? Since we have already deprecated this field and the field is optional, I dont think this will break them. Better to check

PR #1763, opened 2023-09-25
Migrate to centralised reusable workflows

TharmiganK:
## Purpose

> $Subject

Part of https://github.com/ballerina-platform/ballerina-standard-library/issues/4778

## Examples

N/A

## Checklist
- [ ] Linked to an issue
- [ ] ~Updated the changelog~
- [ ] ~Added tests~
- [ ] ~Updated the spec~


MohamedSabthar (approved):

.github/workflows/pull-request.yml
```diff
@@ -3,73 +3,7 @@ name: Ballerina HTTP module build
 on: [pull_request]
 
 jobs:
-  ubuntu-build:
```
MohamedSabthar, line 6: The workflow name should be: `PR Build`

.github/workflows/pull-request.yml
```diff
@@ -3,73 +3,7 @@ name: Ballerina HTTP module build
...
-            uses: actions/setup-java@v3
-            with:
-                distribution: 'temurin'
-                java-version: 17.0.7
-          - name: Build with Gradle
-            env:
-                packageUser: ${{ github.actor }}
-                packagePAT: ${{ secrets.GITHUB_TOKEN }}
-            run: ./gradlew build -x http-native:test
+  call_workflow:
```
MohamedSabthar, line 6: Don't we have a to skip `-x http-native:test` here ?

TharmiganK: We have an additional step to run Ubuntu tests without native tests.
How to we handle it here? Should I add another job to that?

.github/workflows/pull-request.yml
TharmiganK: Ack, will change

.github/workflows/pull-request.yml
MohamedSabthar: Ok. If there is no need to separate this step in to two, We don't have to add the skip flag here.

TharmiganK: Ack, so we will be having two builds and there will be no `Build on Ubuntu without native tests` right?

.github/workflows/pull-request.yml
TharmiganK: Fixed via [a773422](https://github.com/ballerina-platform/module-ballerina-http/pull/1763/commits/a77342294fbc3308d953608063d9027ce3b21d6a)

MohamedSabthar (approved):

PR #1841, opened 2024-01-17
Expose evictable idle time for a connection

dilanSachi:
## Purpose
$subject

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [ ] Added tests - There is already a transport level test for this configuration. Adding a ballerina level test would be much harder
- [ ] Updated the spec

### DIFF
--- a/ballerina/http_client_connection_pool.bal
+++ b/ballerina/http_client_connection_pool.bal
@@ -27,11 +27,13 @@ configurable int maxActiveStreamsPerConnection = 100;
 # + maxIdleConnections - Maximum number of idle connections allowed per pool.
 # + waitTime - Maximum amount of time (in seconds), the client should wait for an idle connection before it sends an error when the pool is exhausted
 # + maxActiveStreamsPerConnection - Maximum active streams per connection. This only applies to HTTP/2. Default value is 100
+# + evictableIdleTime - Evictable time for an idle connection in seconds. Default value is 5 minutes
 public type PoolConfiguration record {|
     int maxActiveConnections = maxActiveConnections;
     int maxIdleConnections = maxIdleConnections;
     decimal waitTime = waitTime;
     int maxActiveStreamsPerConnection = maxActiveStreamsPerConnection;
+    int evictableIdleTime = 5 * 60;
 |};
 //This is a hack to get the global map initialized, without involving locking.
 class ConnectionManager {
--- a/native/src/main/java/io/ballerina/stdlib/http/api/HttpUtil.java
+++ b/native/src/main/java/io/ballerina/stdlib/http/api/HttpUtil.java
@@ -1352,6 +1352,10 @@ public class HttpUtil {
                 maxActiveStreamsPerConnection == -1 ? Integer.MAX_VALUE : validateConfig(
                         maxActiveStreamsPerConnection,
                         HttpConstants.CONNECTION_POOLING_MAX_ACTIVE_STREAMS_PER_CONNECTION.getValue()));
+
+        long minEvictableIdleTime =
+                poolRecord.getIntValue(HttpConstants.CONNECTION_POOLING_EVICTABLE_IDLE_TIME);
+        poolConfiguration.setMinEvictableIdleTime(minEvictableIdleTime);
     }
 
     private static int validateConfig(long value, String configName) {
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,11 @@ This file contains all the notable changes done to the Ballerina HTTP package th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to 
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+- [Expose `evictableIdleTime` for a HTTP connection](https://github.com/ballerina-platform/ballerina-library/issues/5951)
+
 ## [2.10.5] - 2023-12-06
 
 ### Fixed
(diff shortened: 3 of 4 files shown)


ballerina/http_client_connection_pool.bal
```diff
@@ -27,11 +27,13 @@ configurable int maxActiveStreamsPerConnection = 100;
 # + maxIdleConnections - Maximum number of idle connections allowed per pool.
 # + waitTime - Maximum amount of time (in seconds), the client should wait for an idle connection before it sends an error when the pool is exhausted
 # + maxActiveStreamsPerConnection - Maximum active streams per connection. This only applies to HTTP/2. Default value is 100
+# + evictableIdleTime - Evictable time for an idle connection in seconds. Default value is 5 minutes
```
TharmiganK, line 30: This sets the minimum value for evictable idle time of the pool, so isn't it better to name this as:
- `evictableIdleTimeout` or
- `minEvictableIdleTime`

native/src/main/java/io/ballerina/stdlib/http/api/HttpUtil.java
```diff
@@ -1352,6 +1352,10 @@ public static void populatePoolingConfig(BMap poolRecord, PoolConfiguration pool
                 maxActiveStreamsPerConnection == -1 ? Integer.MAX_VALUE : validateConfig(
                         maxActiveStreamsPerConnection,
                         HttpConstants.CONNECTION_POOLING_MAX_ACTIVE_STREAMS_PER_CONNECTION.getValue()));
+
+        long minEvictableIdleTime =
+                poolRecord.getIntValue(HttpConstants.CONNECTION_POOLING_EVICTABLE_IDLE_TIME);
+        poolConfiguration.setMinEvictableIdleTime(minEvictableIdleTime);
```
TharmiganK, line 1358: Isn't this function sets the minimum time as milliseconds?

ballerina/http_client_connection_pool.bal
```diff
@@ -27,11 +27,13 @@ configurable int maxActiveStreamsPerConnection = 100;
...
 public type PoolConfiguration record {|
     int maxActiveConnections = maxActiveConnections;
     int maxIdleConnections = maxIdleConnections;
     decimal waitTime = waitTime;
     int maxActiveStreamsPerConnection = maxActiveStreamsPerConnection;
+    int evictableIdleTime = 5 * 60;
```
TharmiganK, line 36: We can directly put 300 here right?
```suggestion
    int evictableIdleTime = 300;
```

ballerina/http_client_connection_pool.bal
```diff
@@ -27,11 +27,13 @@ configurable int maxActiveStreamsPerConnection = 100;
 # + maxIdleConnections - Maximum number of idle connections allowed per pool.
 # + waitTime - Maximum amount of time (in seconds), the client should wait for an idle connection before it sends an error when the pool is exhausted
 # + maxActiveStreamsPerConnection - Maximum active streams per connection. This only applies to HTTP/2. Default value is 100
+# + evictableIdleTime - Evictable time for an idle connection in seconds. Default value is 5 minutes
 public type PoolConfiguration record {|
```
TharmiganK, line 31: There is another configuration for the eviction interval right? - `timeBetweenEvictionRuns`
The default value for this configuration is 30 seconds. Shouldn't we expose that as well?
@shafreenAnfar WDYT?

ballerina/http_client_connection_pool.bal
TharmiganK, line 36: Since we are configuring this in seconds in ballerina and Java configuration is in milliseconds. Better to set it as a `decimal` value.

native/src/main/java/io/ballerina/stdlib/http/api/HttpUtil.java
```diff
@@ -1352,6 +1352,10 @@ public static void populatePoolingConfig(BMap poolRecord, PoolConfiguration pool
                 maxActiveStreamsPerConnection == -1 ? Integer.MAX_VALUE : validateConfig(
                         maxActiveStreamsPerConnection,
                         HttpConstants.CONNECTION_POOLING_MAX_ACTIVE_STREAMS_PER_CONNECTION.getValue()));
+
+        long minEvictableIdleTime =
+                poolRecord.getIntValue(HttpConstants.CONNECTION_POOLING_EVICTABLE_IDLE_TIME);
```
TharmiganK, line 1357: We need to handle the negative value here right?

ballerina/http_client_connection_pool.bal
dilanSachi: Yeah. Felt that this would be easier for a user to identify its 5 minutes

dilanSachi: Fixed in https://github.com/ballerina-platform/module-ballerina-http/pull/1841/commits/1b88b37adec5c89f232aa6b229cbfaa07613f4e1

native/src/main/java/io/ballerina/stdlib/http/api/HttpUtil.java
dilanSachi: Fixed in https://github.com/ballerina-platform/module-ballerina-http/pull/1841/commits/1b88b37adec5c89f232aa6b229cbfaa07613f4e1

ballerina/http_client_connection_pool.bal
dilanSachi: Fixed in https://github.com/ballerina-platform/module-ballerina-http/pull/1841/commits/1b88b37adec5c89f232aa6b229cbfaa07613f4e1

ballerina/http_client_connection_pool.bal
TharmiganK: Yes, that's fine. But since we are anyway adding the default value in the doc comment, I don't think we need to give the value like this

dilanSachi: ack

ballerina/http_client_connection_pool.bal
dilanSachi: Updated

TharmiganK (approved):
LGTM. Can we check the spec as well? If we have a reference to this PoolConfiguration in the spec, then we need to update that as well

dilanSachi:
> LGTM. Can we check the spec as well? If we have a reference to this PoolConfiguration in the spec, then we need to update that as well

We only have the `PoolConfiguration` mentioned in `ClientConfiguration`. There is no explicit record definition

PR #1875, opened 2024-03-01
Change default TLS named groups

hwupathum:
## Purpose

Related issues
-  https://github.com/ballerina-platform/ballerina-lang/issues/42345
- https://github.com/ballerina-platform/ballerina-library/issues/6200

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec


native/src/main/java/io/ballerina/stdlib/http/api/HttpUtil.java
```diff
@@ -1845,6 +1848,9 @@ private static void evaluateProtocolField(BMap<BString, Object> protocol,
         if (!sslProtocol.isBlank()) {
             sslConfiguration.setSSLProtocol(sslProtocol);
         }
+        if (System.getProperty(JAVA_CONFIG_TLS_NAMED_GROUPS).isBlank()) {
```
TharmiganK, line 1851: We have to do a null check before checking `isBlank()`. Also it should be enough to check only for null property right? Why do we need to check whether it is blank?

Bhashinee:
Shall we update the changelog?

Bhashinee (approved):

PR #1882, opened 2024-03-05
[2201.6.x] Fix overwriting the Host header

TharmiganK:
## Purpose

> Related to: https://github.com/ballerina-platform/ballerina-library/issues/6133

**Note:** With this improvement users can override the `Host` header, but **this does not change the connection internals**. We will still connect to the endpoint defined in the client `URL` but the `Host` header send in the request can be overridable. This is in line with the cURL client.

For HTTP/2, we expose the `Host` header from the pseudo header. Since the pseudo header exposure is added from `2201.7.x`, we cannot add test with HTTP/2 in `2201.6.x`. For the sake of completeness, I have back ported the pseudo headers fix with this PR
> Related to: https://github.com/ballerina-platform/ballerina-library/issues/4732

## Examples

N/A

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] ~Updated the spec~

### DIFF
--- a/ballerina-tests/http2-tests/tests/http2_service_dispatching_header_param_binding_test.bal
+++ b/ballerina-tests/http2-tests/tests/http2_service_dispatching_header_param_binding_test.bal
@@ -565,7 +565,7 @@ function testHeaderRecordParamOfPureTypeHeadersNegative() returns error? {
     });
     if response is http:ClientRequestError {
         test:assertEquals(response.detail().statusCode, 400, msg = "Found unexpected output");
-        common:assertErrorHeaderValue(response.detail().headers[common:CONTENT_TYPE], common:APPLICATION_JSON);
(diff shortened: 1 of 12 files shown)


ballerina-tests/http2-tests/tests/http2_service_dispatching_header_param_binding_test.bal
```diff
@@ -0,0 +1,838 @@
+// Copyright (c) 2023, WSO2 LLC. (http://www.wso2.org) All Rights Reserved.
```
TharmiganK, line 1:
```suggestion
// Copyright (c) 2024, WSO2 LLC. (http://www.wso2.org).
```

dilanSachi (approved):

PR #1954, opened 2024-04-19
Fix hanging issue in the HTTP client after status code binding feature

TharmiganK:
## Purpose

> $Subject

Related to https://github.com/ballerina-platform/ballerina-library/issues/6356

Here, we have used a hack to avoid the countdown latch and used the bal future to return the result from the ballerina call directly. With this there is an **additional call to Java in order to build the status code response** (since the status code response is not anydata, we cannot create it using cloneWithType). So we have decided to add this new feature with a new client: `http:StatusCodeClient` and the behaviour of the `http:Client` will remain same as in Update 8.

## Examples

N/A

## Checklist
- [x] Linked to an issue
- [ ] ~Updated the changelog~
- [ ] ~Added tests~
- [ ] ~Updated the spec~
- [ ] Checked the impact on OpenAPI generation : Need to address this!


TharmiganK:
Will send a different PR to improve code coverage with the new client

shafreenAnfar (approved):

PR #1987, opened 2024-05-07
[5.x] Expose http connection eviction configurations in the client level

dilanSachi:
## Purpose
$subject
Related to https://github.com/ballerina-platform/ballerina-library/issues/6503
## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec

### DIFF
--- a/native/src/main/java/io/ballerina/stdlib/http/api/HttpUtil.java
+++ b/native/src/main/java/io/ballerina/stdlib/http/api/HttpUtil.java
@@ -1344,6 +1344,14 @@ public class HttpUtil {
                 maxActiveStreamsPerConnection == -1 ? Integer.MAX_VALUE : validateConfig(
                         maxActiveStreamsPerConnection,
                         HttpConstants.CONNECTION_POOLING_MAX_ACTIVE_STREAMS_PER_CONNECTION.getValue()));
+        double minEvictableIdleTime =
+                ((BDecimal) poolRecord.get(HttpConstants.CONNECTION_POOLING_EVICTABLE_IDLE_TIME)).floatValue();
+        poolConfiguration.setMinEvictableIdleTime(minEvictableIdleTime < 0 ? 0 : (long) minEvictableIdleTime * 1000);
+
+        double timeBetweenEvictionRuns =
+                ((BDecimal) poolRecord.get(HttpConstants.CONNECTION_POOLING_TIME_BETWEEN_EVICTION_RUNS)).floatValue();
+        poolConfiguration.setTimeBetweenEvictionRuns(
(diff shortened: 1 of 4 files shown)


native/src/main/java/io/ballerina/stdlib/http/api/HttpUtil.java
```diff
@@ -1344,6 +1344,14 @@ public static void populatePoolingConfig(BMap poolRecord, PoolConfiguration pool
                 maxActiveStreamsPerConnection == -1 ? Integer.MAX_VALUE : validateConfig(
                         maxActiveStreamsPerConnection,
                         HttpConstants.CONNECTION_POOLING_MAX_ACTIVE_STREAMS_PER_CONNECTION.getValue()));
+        double minEvictableIdleTime =
+                ((BDecimal) poolRecord.get(HttpConstants.CONNECTION_POOLING_EVICTABLE_IDLE_TIME)).floatValue();
+        poolConfiguration.setMinEvictableIdleTime(minEvictableIdleTime < 0 ? 0 : (long) minEvictableIdleTime * 1000);
```
ayeshLK, line 1349: Usually if a configuration value has a negative value like -1 ect. the configuration will be considered as disabled. Can we apply that principle here rather that setting the value as 0 ?

Bhashinee: I think this is because, here we are setting this value to the Apache connection pool which is a third-party library, and there "0" is considered as the default.

dilanSachi: This is a patch backported from 2201.8.x. I dont think its good to change the configuration behaviour.
And also, this is a value we forward to apache generic object pool. so disabling such config does not make sense IMO

dilanSachi: This is a patch backported from 2201.8.x. I dont think its good to change the configuration behaviour.
And also, this is a value we forward to apache generic object pool. so disabling such config does not make sense IMO

TharmiganK (approved):
LGTM

Bhashinee (approved):

native/src/main/java/io/ballerina/stdlib/http/api/HttpUtil.java
ayeshLK: Ack. But, I think our original approach should have used the disabling config when the value is a negative value. But, since you are passing this config to Apache generic pool object it might be doing something similar from within. But, its better to confirm the behaviour when the value is negative or zero.

PR #2066, opened 2024-07-10
Check stale connection eviction logic to listener side

dilanSachi:
## Purpose
$subject
Fixes https://github.com/ballerina-platform/ballerina-library/issues/6734

Checked a scenario as below.
* Http2 client creates 100 connections per seconds.
* For each connection, once a single communication is done, the client sends a GO_AWAY, but does not close the connection and keeps it open.
* In the existing way, the number of connections keeps on increasing.
* With this change, when the stale timeout happens, the related connections gets closed. Therefore, after sometime, the number of open connections gets to a fixed amount and does not increase.
* This is the output of `while true; do lsof -i tcp -a -p 222685 | wc -l; sleep 2; done 2>/dev/null` with single run of the above scenario.
```
2
2
2
202
402
602
802
1002
1102
1106
1200
1201
1201
1201
1200
1200
1200
1201
1173
1180
1202
1180
1138
1101
1101
1100
1098
1099
1100
1102
1100
1098
1097
2
2
2
2
```
* Following is the existing version's output.
```
2
2
2
2
202
402
602
802
1002
1202
1402
1602
1802
2002
2202
2402
2602
2802
3002
3202
3402
3702
3902
4102
4302
4502
4702
4902
5102
2
2
2
2
2
2
2
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec

Ran a  load test as well. Results can be seen here - https://github.com/ballerina-platform/module-ballerina-http/pull/2080/files
Looks like no issues.


changelog.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Added
 
 - [Introduce default status code response record](https://github.com/ballerina-platform/ballerina-library/issues/6491)
+- [Added connection eviction feature to handle connections that receive GO_AWAY from the client](https://github.com/ballerina-platform/ballerina-library/issues/6734)
```
TharmiganK, line 13:
```suggestion
- [Add connection eviction feature to handle connections that receive GO_AWAY from the client](https://github.com/ballerina-platform/ballerina-library/issues/6734)
```

TharmiganK (approved):
LGTM

PR #2108, opened 2024-08-14
Update HTTP specification

TharmiganK:
## Purpose

> $Subject

Fixes: https://github.com/ballerina-platform/ballerina-library/issues/6816
Fixes: https://github.com/ballerina-platform/ballerina-library/issues/6289

## Examples

N/A

## Checklist
- [x] Linked to an issue
- [ ] ~Updated the changelog~
- [ ] ~Added tests~
- [ ] ~Updated the spec~


docs/spec/spec.md
```diff
@@ -294,14 +297,68 @@ public function main() {
...
+                }
+            } Album album) returns Album|ErrorPayloadBadRequest;
+};
+```
+
+The service contract type can have the annotations which are supported on the service declaration. Additionally, a 
+`basePath` can be defined in the `ServiceConfig` annotation to define the base path of the service. This field is 
+optional and if not defined, the base path will be defaulted to `/`.
+
+Following is an example of a service implemented with the above service contract type:
```
lnash94, line 334:
```suggestion
Following is an example of a service implemented with the above service contract type:

```

docs/spec/spec.md
```diff
@@ -294,14 +297,68 @@ public function main() {
...
 ```
 
+### 2.2.6. Service contract type
+
+The service contract type is a distinct type which is used to represent the service contract. This service contract
+type can be used along with the service declaration to implement a service which ensures that the service is compliant
+with the contract. Additionally, all the metadata related to the service can be defined within the service contract 
+type which makes the service declaration clean and readable.
+
+Following is an example of a service contract type:
```
lnash94, line 307:
```suggestion
Following is an example of a service contract type:

```

lnash94 (approved):

PR #2135, opened 2024-09-03
Add `anydata` support for `setPayload` methods

TharmiganK:
## Purpose

> $Subject

Fixes: https://github.com/ballerina-platform/ballerina-library/issues/6954

The `setPayload` with `anydata` uses the `setJsonPayload` with the `toJson` value of the `anydata`.

## Example

```bal
// Open record
type AnydataRecord record {
    string a;
    int b;
};

AnydataRecord testValue = {a: "ballerina", b: 1};

// Request Object
http:Request req = new;
req.setPayload(testValue);

// Response Object
http:Response res = new;
res.setPayload(testValue);
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] ~Updated the spec~
- [ ] [Checked breaking changes with dependent modules](https://github.com/ballerina-platform/module-ballerina-http/actions/runs/10695077318)


daneshk (approved):
LGTM

ballerina/http_response.bal
```diff
@@ -517,6 +517,8 @@ public class Response {
             self.setBodyParts(payload);
         } else if payload is stream<SseEvent, error?> {
             self.setSseEventStream(payload);
+        } else if payload is anydata {
+            self.setJsonPayload(payload.toJson());
```
SasinduDilshara, line 521: import ballerina/io;

public function main() {
    anydata a = re `a`;
    io:println(a.toJson());
}

If this kind of a scenario, it will throws a runtime error of 
```
error: {ballerina/lang.value}ConversionError {"message":"'anydata & readonly' value cannot be converted to 'json'"}
        at ballerina.lang.value.0:toJson(value.bal:262)
           admin.issue_6926.0:main(main.bal:6)
```

TharmiganK: You mean it will panic right? Shouldn't we handle this in the lang level. The `toJson` returns a `json` not `json|error`.

SasinduDilshara: +1, can you create a issue for this, As a work around can you create a type check for this `if payload is string:RegExp`

TharmiganK: I think this is a very rare case (providing a regular expression in the `setPayload` method). So I don't think we need to explicitly handle this. Since this is a lang issue, if they decided to fix it by giving an error or string value of the regular expression, we will change accordingly in the future. 

Created an issue here: https://github.com/ballerina-platform/ballerina-lang/issues/43360

SasinduDilshara (dismissed):

ballerina/http_response.bal
```diff
@@ -517,6 +517,8 @@ public class Response {
             self.setBodyParts(payload);
         } else if payload is stream<SseEvent, error?> {
             self.setSseEventStream(payload);
+        } else if payload is anydata {
```
MaryamZi, line 520: toJson will always create a new value even if the value is already json. May make sense to add an is json check and handle that separately.

ballerina/http_request.bal
```diff
@@ -514,11 +514,12 @@ public class Request {
     # as an optional parameter. If the content type parameter is not provided then the default value derived
     # from the payload will be used as content-type only when there are no existing content-type header.
     #
-    # + payload - Payload can be of type `string`, `xml`, `json`, `byte[]`, `stream<byte[], io:Error?>`
+    # + payload - Payload can be of type `anydata`, `stream<byte[], io:Error?>`
```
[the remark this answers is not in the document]
TharmiganK: Will mention this 👍🏼

ballerina/http_response.bal
TharmiganK: Yes, that is handled before this check: https://github.com/ballerina-platform/module-ballerina-http/blob/0cb5cf9e14df0023d45b45636d16c2b4c9864652/ballerina/http_response.bal#L512

ballerina/http_request.bal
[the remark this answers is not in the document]
TharmiganK: Fixed via [0c9bbbe](https://github.com/ballerina-platform/module-ballerina-http/pull/2135/commits/0c9bbbe9648f36d30e2804bdb63b41a5d9e7f866)

ballerina/http_request.bal
```diff
@@ -512,13 +512,15 @@ public class Request {
 
     # Sets the request payload. This method overrides any existing content-type by passing the content-type
     # as an optional parameter. If the content type parameter is not provided then the default value derived
-    # from the payload will be used as content-type only when there are no existing content-type header.
+    # from the payload will be used as content-type only when there are no existing content-type header. If
+    # the payload is non-json typed value then the value is converted to json using the `toJson` method.
```
MaryamZi, line 516: Not entirely an accurate description for xml, right?

ballerina/http_request.bal
```diff
@@ -512,13 +512,15 @@ public class Request {
...
     #
-    # + payload - Payload can be of type `string`, `xml`, `json`, `byte[]`, `stream<byte[], io:Error?>`
+    # + payload - Payload can be of type `anydata`, `stream<byte[], io:Error?>`
```
MaryamZi, line 518: Would we have to explicitly mention xml, json, etc. as examples? Just thinking of someone getting started with Ballerina.

TharmiganK: We usually don't suggest users to use request/response objects with setPayload methods unless it is an advanced scenario like multipart. May be we can include a more detailed description in the spec

ballerina/http_request.bal
TharmiganK: Yes, will update it

TharmiganK: Updated via https://github.com/ballerina-platform/module-ballerina-http/pull/2135

lnash94 (approved):

PR #2143, opened 2024-09-09
[master] Add Query Parameter Annotation Mapping 

lnash94:
## Purpose
Fixes: https://github.com/ballerina-platform/ballerina-library/issues/6983

Partial fix for the main issue: https://github.com/ballerina-platform/ballerina-library/issues/6747

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] Updated the spec https://github.com/ballerina-platform/ballerina-library/issues/6987
- [x] Checked the impact on OpenAPI generation - https://github.com/ballerina-platform/ballerina-library/issues/6984


ballerina-tests/http-client-tests/tests/client_resource_method_tests.bal
```diff
@@ -280,3 +295,35 @@ function testClientResourceWithBasicRestType() returns error? {
...
+
+    Person person = {
+        name: "Harry",
+        age: 29
+    };
+
+    response = check clientResourceMethodsClientEP->/query/bar/[99].post( person, params = queries);
+    test:assertEquals(response, "Greetings! Mr. Jhon Harry");
+
+}
```
NipunaRanasinghe, line 329:
```suggestion
    test:assertEquals(response, "Greetings! Mr. Jhon Harry");
}
```

ballerina-tests/http-client-tests/tests/client_resource_method_tests.bal
```diff
@@ -120,6 +120,21 @@ service on clientResourceMethodsServerEP {
     }
 }
 
+service /query on clientResourceMethodsServerEP {
+
+    resource function get bar(string first\-name, string 'table, string age) returns string {
+        return "Here are query params: " + first\-name + "," +'table + "," + age.toString();
```
SasinduDilshara, line 126: Shall we use string concatenation here.

ballerina-tests/http-client-tests/tests/client_resource_method_tests.bal
```diff
@@ -120,6 +120,21 @@ service on clientResourceMethodsServerEP {
...
+    }
+
+    resource function get bar/[int pathParam](string first\-name, string 'table, string age) returns string {
+        return "Here are path param with query params: " + first\-name + "," +'table + "," + age.toString();
```
SasinduDilshara, line 130:
```suggestion
        return "Here are path param with query params: " + first\-name + "," +'table + "," + age;
```

ballerina-tests/http-client-tests/tests/client_resource_method_tests.bal
```diff
@@ -120,6 +120,21 @@ service on clientResourceMethodsServerEP {
...
+    }
+
+    resource function post bar/[int pathParam](Person body, string first\-name, string 'table, string age) returns string {
+        return "Greetings! Mr. " + first\-name + " "  + body.name ;
```
SasinduDilshara, line 134:
```suggestion
        return "Greetings! Mr. " + first\-name + " "  + body.name;
```

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -293,22 +294,49 @@ private static String[] getPathStringArray(BArray pathArray) {
...
     }
 
+    /**
+     * This util function extracts the query name with the query annotation.
+     *
+     * @param params - Parameter map
+     * @return Map of string with overridden query param names
+     */
+    private static Map<String, String> getQueryNameMapping(BMap params) {
+        RecordType queryRecord = (RecordType) params.getType();
```
TharmiganK, line 323: The type can be a map also. So we nee to check the type here before casting

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -293,22 +294,49 @@ private static String[] getPathStringArray(BArray pathArray) {
...
+        BMap<BString, Object> annotations = queryRecord.getAnnotations();
+        Map<String, String> annotationValues = new HashMap<>();
+
+        for (Map.Entry<BString, Object> annotation: annotations.entrySet()) {
+            //store query parameters real name : $field$.<query name>
+            BString key = annotation.getKey().substring(8, annotation.getKey().length());
```
TharmiganK, line 329: Better to split it with the regex(starts with `$field$.`) rather than specific length, then it will be easier to change if the lang did any changes in the internal representation.

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -293,22 +294,49 @@ private static String[] getPathStringArray(BArray pathArray) {
...
+            BMap value = (BMap) annotation.getValue();
+            for (Object object : value.values()) {
```
TharmiganK, line 331: We need to check for the specific annotation: `http:Query` rather than reading the `name` field from one of the annotation. We can have multiple annotations on a field which can have the `name` field. Better to add a test case as well

lnash94: +1 , will change the logic

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
lnash94: Addressed via https://github.com/ballerina-platform/module-ballerina-http/pull/2143/commits/93a28143565d062f1a2a34e36bbfda35ea6e22a4

ballerina/http_annotation.bal
```diff
@@ -115,10 +115,14 @@ public type HttpHeader record {|
 public annotation HttpHeader Header on parameter;
 
 # Defines the query resource signature parameter.
-public type HttpQuery record {||};
+# 
+# + name - Specifies the name of the required query
```
TharmiganK, line 119:
```suggestion
# + name - Specifies the name of the query parameter
```

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -293,22 +296,66 @@ private static String[] getPathStringArray(BArray pathArray) {
...
+        if (!(params.getType() instanceof RecordType queryRecord)) {
+            return annotationValues;
+        }
+        BMap<BString, Object> annotations = queryRecord.getAnnotations();
+
+        for (Map.Entry<BString, Object> annotation: annotations.entrySet()) {
+            BMap value = (BMap) annotation.getValue();
+            Object[] keys = value.getKeys();
+            for (Object annotRef: keys) {
+                String refRegex = "^ballerina/http:\\d+:(Query)$";
```
TharmiganK, line 335: You could use the following rather than the string value: `ModuleUtils.getHttpPackageIdentifier() + COLON + "Query"`

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -293,22 +296,66 @@ private static String[] getPathStringArray(BArray pathArray) {
...
+
+    private static void extractedFieldName(Map<String, String> annotationValues, Map.Entry<BString, Object> annotation,
+                                           BMap value) {
+        for (Object object : value.values()) {
+            //store query parameters real name : $field$.<query name>
+            String regex = "(\\$field\\$\\.)";
+            String[] parts = Pattern.compile(regex).split(annotation.getKey().getValue());
+            String key = parts[1];
+            BMap bMap = (BMap) object;
+            BString overrideName = (BString) bMap.get(StringUtils.fromString("name"));
```
TharmiganK, line 354: You can use the existing constant:
```suggestion
            BString overrideName = (BString) bMap.get(HttpConstants.ANN_FIELD_NAME);
```

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -293,22 +296,66 @@ private static String[] getPathStringArray(BArray pathArray) {
...
+        return annotationValues;
+    }
+
+    private static void extractedFieldName(Map<String, String> annotationValues, Map.Entry<BString, Object> annotation,
+                                           BMap value) {
+        for (Object object : value.values()) {
+            //store query parameters real name : $field$.<query name>
+            String regex = "(\\$field\\$\\.)";
+            String[] parts = Pattern.compile(regex).split(annotation.getKey().getValue());
+            String key = parts[1];
```
TharmiganK, line 352:
```suggestion
            String fieldName = parts[1];
```

Does the field name is properly unescaped or do we need to manually do that?

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -293,22 +296,66 @@ private static String[] getPathStringArray(BArray pathArray) {
...
+            }
+        }
+        return annotationValues;
+    }
+
+    private static void extractedFieldName(Map<String, String> annotationValues, Map.Entry<BString, Object> annotation,
+                                           BMap value) {
+        for (Object object : value.values()) {
+            //store query parameters real name : $field$.<query name>
+            String regex = "(\\$field\\$\\.)";
```
TharmiganK, line 350: Define a constant in the class level

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
TharmiganK, line 354: Since the value is optional, better to check whether the value is `BString` before casting. Even though it is not 80% case, you can add a `nil` value for the name:
```bal
public type QueryParams record {|
    @http:Query {name: ()}
    string firstName;
|};
```

lnash94: Added the type check before casting, and curious if it allows a `()` value in the compilation although the field mentions having `string`  instead of `string?`.

lnash94: If the user provides an empty name value, such as `()` or `""`, the current implementation will fall back to the field name. Should we add an error or warning message prompting the user to provide a valid name, or should we keep the current behavior where an error response is returned?

TharmiganK: +1 to fallback the field name. There is no need for error or warning since this is not 80% case

lnash94: address via https://github.com/ballerina-platform/module-ballerina-http/pull/2143/commits/1218d3d373e183282f02752cd48815992384c3f6 and this for keeping the field name when the name is absent https://github.com/ballerina-platform/module-ballerina-http/blob/1218d3d373e183282f02752cd48815992384c3f6/native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java#L311

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -293,22 +300,67 @@ private static String[] getPathStringArray(BArray pathArray) {
...
+        if (!(params.getType() instanceof RecordType queryRecord)) {
+            return annotationValues;
+        }
+        BMap<BString, Object> queryFields = queryRecord.getAnnotations();
+
+        for (Map.Entry<BString, Object> qField: queryFields.entrySet()) {
+            BMap value = (BMap) qField.getValue();
+            Object[] keys = value.getKeys();
+            for (Object annotRef: keys) {
+                String refRegex = ModuleUtils.getHttpPackageIdentifier() + COLON + "Query";
```
TharmiganK, line 339: Don't we have a constant for "Query"?

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -293,22 +300,67 @@ private static String[] getPathStringArray(BArray pathArray) {
...
+    private static void extractedFieldName(Map<String, String> annotationValues, Map.Entry<BString, Object> qField,
+                                           BMap value) {
+        String[] parts = Pattern.compile(REGEX_FOR_FIELD).split(qField.getKey().getValue());
+        String fieldName = unescapeIdentifier(parts[1]);
+        Object overrideValue = value.get(HttpConstants.ANN_FIELD_NAME);
+        if (!(overrideValue instanceof BString overrideName)) {
+            return;
+        }
+        annotationValues.put(fieldName, overrideName.getValue());
+    }
```
TharmiganK, line 360:
```suggestion
    }

```

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -293,22 +300,67 @@ private static String[] getPathStringArray(BArray pathArray) {
...
+    public static String unescapeIdentifier(String parameterName) {
+        return parameterName.replaceAll(ESCAPE_SLASH, EMPTY);
+    }
```
TharmiganK, line 363:
```suggestion
    }

```

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -293,22 +301,69 @@ private static String[] getPathStringArray(BArray pathArray) {
...
+    /**
+     * This util function extracts the query name with the query annotation.
+     *
+     * @param params - Parameter map
+     * @return Map of string with overridden query param names
+     */
+    private static Map<String, String> getQueryNameMapping(BMap params) {
+        Map<String, String> annotationValues = new HashMap<>();
+        if (!(params.getType() instanceof RecordType queryRecord)) {
+            return annotationValues;
```
TharmiganK, line 332: Shall we add a test case to cover this part of the code? I think we can try to pass a map for queries

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -293,22 +301,69 @@ private static String[] getPathStringArray(BArray pathArray) {
...
+        return annotationValues;
+    }
+
+    private static void extractedFieldName(Map<String, String> annotationValues, Map.Entry<BString, Object> qField,
+                                           BMap value) {
+        String[] parts = Pattern.compile(REGEX_FOR_FIELD).split(qField.getKey().getValue());
+        String fieldName = unescapeIdentifier(parts[1]);
+        Object overrideValue = value.get(HttpConstants.ANN_FIELD_NAME);
+        if (!(overrideValue instanceof BString overrideName)) {
+            return;
```
TharmiganK, line 358: Shall we add a test case where we set the `name: ()` and another test case without the name field?

lnash94: Added via bb44773

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
lnash94: Removed this check since we only allow the query with the QueryParam type.

TharmiganK (approved):
LGTM

PR #2202, opened 2024-10-28
Add static code analyzer rules

MohamedSabthar:
## Purpose
$Subject
Related issue: https://github.com/ballerina-platform/ballerina-library/issues/7283

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] Updated the spec
- [x] Checked the impact on OpenAPI generation


compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/StaticCodeAnalyzerTest.java
```diff
@@ -0,0 +1,120 @@
...
+            .get("../", "target", "ballerina-runtime", "bin", "bal").toAbsolutePath();
+    private static final Path JSON_RULES_FILE_PATH = Paths
+            .get("../", "compiler-plugin", "src", "main", "resources", "rules.json").toAbsolutePath();
+    private static final String SCAN_COMMAND = "scan";
+
+    @BeforeSuite
+    public void pullScanTool() throws IOException, InterruptedException {
+        String scanToolVersion = "0.1.0";
+        ProcessBuilder processBuilder = new ProcessBuilder(BALLERINA_PATH.toString(),
+                "tool", "pull", SCAN_COMMAND + ":" + scanToolVersion, "--repository=local");
```
MohamedSabthar, line 56: This line will be updated to
```
ProcessBuilder processBuilder = new ProcessBuilder(BALLERINA_PATH.toString(),
                "tool", "pull", SCAN_COMMAND);
```
 after releasing the scan tool command

MohamedSabthar: updated

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpServiceAnalyzer.java
```diff
@@ -0,0 +1,46 @@
...
+    private final Reporter reporter;
+
+    public HttpServiceAnalyzer(Reporter reporter) {
+        this.reporter = reporter;
+    }
+
+    public void validateServiceMembers(NodeList<Node> members, Document document) {
+        // TODO: fix location, currently getting always -1 than expected
+        HttpCompilerPluginUtil.getResourceMethodWithDefaultAccessor(members).forEach(definition -> {
+            Location accessorLocation = definition.functionName().location();
```
TharmiganK, line 42: Can't we get the entire node location of the resource: `definition.location()`?
Otherwise we need to get the function definition node to get the location of the `default` keyword

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpServiceClassAnalyzer.java
```diff
@@ -0,0 +1,41 @@
...
+
+class HttpServiceClassAnalyzer extends HttpServiceAnalyzer {
+    public HttpServiceClassAnalyzer(Reporter reporter) {
+        super(reporter);
+    }
+
+    @Override
+    public void perform(SyntaxNodeAnalysisContext context) {
+        ClassDefinitionNode serviceClassDefinitionNode = HttpCompilerPluginUtil.getServiceClassDefinitionNode(context);
+        if (serviceClassDefinitionNode == null) {
```
TharmiganK, line 35:
```suggestion
        if (Objects.isNull(serviceClassDefinitionNode)) {
```

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpServiceDeclarationAnalyzer.java
```diff
@@ -0,0 +1,41 @@
...
+
+package io.ballerina.stdlib.http.compiler.staticcodeanalyzer;
+
+import io.ballerina.compiler.syntax.tree.ServiceDeclarationNode;
+import io.ballerina.projects.Document;
+import io.ballerina.projects.plugins.SyntaxNodeAnalysisContext;
+import io.ballerina.scan.Reporter;
+import io.ballerina.stdlib.http.compiler.HttpCompilerPluginUtil;
+
+class HttpServiceDeclarationAnalyzer extends HttpServiceAnalyzer {
```
TharmiganK, line 27: Can't we combine `HttpServiceClassAnalyzer`, `HttpServiceDeclarationAnalyzer` and `HttpServiceObjectTypeAnalyzer` classes? Seems like they have similar code

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/RuleFactory.java
```diff
@@ -0,0 +1,31 @@
...
+
+package io.ballerina.stdlib.http.compiler.staticcodeanalyzer;
+
+import io.ballerina.scan.Rule;
+import io.ballerina.scan.RuleKind;
+
+/**
+ * {@code RuleFactory} contains the logic to create a {@link Rule}.
+ */
+public class RuleFactory {
```
TharmiganK, line 27: Should this be a functionality provided by the ballerina scan package? IIUC we need to have `RuleFactory` and `RuleImpl` in other libraries as well

compiler-plugin/src/main/resources/rules.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "id": 1,
+    "kind": "VULNERABILITY",
+    "description": "Avoid allowing default resource accessor"
```
TharmiganK, line 5: I would like to have a descriptive reason and resolution for these issues. Even though some issues are straight forward, we need to at least mention the reason for marking this as an vulnerability and how we can resolve this issue. Can you check the possibility of configuring them as well?  

For example if you check this Java sonar report: https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=2202&id=ballerina-platform_module-ballerina-http&open=AZLRrEMyf0DC8seLHPIr&tab=code
There is a section called `Why`.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
```diff
@@ -348,6 +362,47 @@ public static ServiceDeclarationNode getServiceDeclarationNode(Node node, Semant
         return serviceDeclarationNode;
     }
 
+    public static ClassDefinitionNode getServiceClassDefinitionNode(SyntaxNodeAnalysisContext ctx) {
+        if (ctx.node().kind() != CLASS_DEFINITION) {
+            return null;
+        }
+        ClassDefinitionNode classDefinitionNode = (ClassDefinitionNode) ctx.node();
+        Optional<Symbol> serviceContractType = ctx.semanticModel().types()
```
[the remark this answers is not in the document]
MohamedSabthar: I see that the http module has both `http:Service` and `http:ServiceContract` as types. Since `Service` objects are also contracts, I was referring to the type of this contract. Since this could be confusing, I'll refactor it for better clarity.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
```diff
@@ -348,6 +362,47 @@ public static ServiceDeclarationNode getServiceDeclarationNode(Node node, Semant
...
+        TypeSymbol serviceType = ((TypeDefinitionSymbol) serviceContractType.get()).typeDescriptor();
+        return classSymbol.subtypeOf(serviceType) ? classDefinitionNode : null;
+    }
+
+    public static AnnotationNode getAnnotationNode(SyntaxNodeAnalysisContext context) {
+        if (context.node().kind() != ANNOTATION) {
+            return null;
+        }
+        Optional<Symbol> symbol = context.semanticModel().symbol(context.node());
+        if (symbol.isPresent()) {
```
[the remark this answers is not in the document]
MohamedSabthar: The early return was intentionally omitted here. I added a workaround code to handle a compiler plugin symbol resolution bug. If we move the workaround code to the beginning (to early return), the subsequent code would become unreachable due to the early return. This approach allows the code to execute normally when there are no symbol resolution issues.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
```diff
@@ -348,6 +362,47 @@ public static ServiceDeclarationNode getServiceDeclarationNode(Node node, Semant
...
+            }
+            return (AnnotationNode) context.node();
+        }
+        // Added as a workaround for: https://github.com/ballerina-platform/ballerina-lang/issues/43525
+        return context.node().toSourceCode().trim().startsWith(AT_TOKEN.stringValue() + HTTP + COLON) ?
+                (AnnotationNode) context.node() : null;
+    }
+
+    private static boolean hasServiceKeyWord(ClassDefinitionNode classDefinitionNode) {
+        return classDefinitionNode.classTypeQualifiers()
```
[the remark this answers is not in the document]
MohamedSabthar: this will always return a `NodeList`. No qualifiers then an empty `NodeList`.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpRule.java
```diff
@@ -0,0 +1,52 @@
...
+import io.ballerina.scan.Rule;
+
+import static io.ballerina.scan.RuleKind.VULNERABILITY;
+import static io.ballerina.stdlib.http.compiler.staticcodeanalyzer.RuleFactory.createRule;
+
+/**
+ * Represents static code rules specific to the Ballerina Http package.
+ */
+public enum HttpRule {
+    AVOID_DEFAULT_RESOURCE_ACCESSOR(createRule(1, "Avoid allowing default resource accessor", VULNERABILITY)),
```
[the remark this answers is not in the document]
MohamedSabthar: Yes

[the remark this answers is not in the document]
MohamedSabthar: Currently the scan tool it self doesn't support a description on how to resolve this issue. And since the description added here will be used as the title in the report page this should not be too long. 

@SasinduDilshara do we have plans to update the format of `rules.json` file ?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpServiceAnalyzer.java
MohamedSabthar: Yes, we can do that. However, we should focus on indicating the accessor location since it makes it easier for users to pinpoint the exact location where they need to make changes.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpServiceClassAnalyzer.java
MohamedSabthar: Since this is procedural code rather than functional, we'll stick to the simple `== null` check. Objects.isNull() is typically preferred in functional-style code.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpRule.java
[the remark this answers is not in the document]
TharmiganK: I meant unique to the package. How does the id is inferred from this number? Is it like `http-1`?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpServiceAnalyzer.java
TharmiganK: Ack, have you tried to get it from the function definition node itself? Otherwise getting a `-1` is a bug right?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpServiceDeclarationAnalyzer.java
MohamedSabthar: Yes, we can achieve that by registering a class like this:
```java
analysisContext.addSyntaxNodeAnalysisTask(new ServiceAnalyzer(reporter), 
    List.of(CLASS_DEFINITION, OBJECT_TYPE_DESC, SERVICE_DECLARATION));
```
However, using a single ServiceAnalyzer would require multiple if-checks in the perform method to handle different syntax kinds. Instead, I preferred using polymorphism over conditional checks and moved common functionality to an abstract HttpServiceAnalyzer class. WDYT?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/RuleFactory.java
MohamedSabthar: The scanner package doesn't provide an API to register rules - this is handled through the `rules.json` file instead. While it's not necessary to have rule-related classes in the library package (we can simply report rules using IDs defined in `rules.json`), using hardcoded IDs reduces code readability. Therefore, I created a Rule interface to obtain rule information, which also helps with testing.

If this not necessary I can remove these classes and update the code with just hardcoded rule ID.

compiler-plugin/src/main/resources/rules.json
MohamedSabthar: +1
@MaryamZi, @SasinduDilshara and I had a chat on this few days back. I believe Scan tool team already have plan to do this.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/RuleFactory.java
TharmiganK: I prefer this `rules.json` approach. It is more readable. Let's have it as it is. 
But the point is we have to duplicate these classes in the other modules. I hope that is fine for now, since we even have an approach to hard-code the values in the code as well

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpRule.java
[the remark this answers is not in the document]
MohamedSabthar: yes it will be prefixed with module name, this will be automatically done by scan tool.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpServiceDeclarationAnalyzer.java
TharmiganK: How about defining an interface to these common types which exposes the common properties? 

I have did something similar for resource function implementation and resource function definition (Interface - `ResourceFunction`, Implementations - `ResourceFunctionDefintion` and `ResourceFunctionDeclaration`)
Can you check whether it is viable here?

compiler-plugin/src/main/resources/rules.json
TharmiganK: Do we have an issue for this?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpServiceAnalyzer.java
MohamedSabthar: When reporting the error, the column/line number is one less than the expected value. This is because the syntax tree location is 0-based. However, when reporting the error, this should be adjusted to a 1-based format. This needs to be fixed in the scan tool itself: https://github.com/ballerina-platform/static-code-analysis-tool/issues/29

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpCompilerPluginUtil.java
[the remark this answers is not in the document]
MohamedSabthar: fixed [2aa962f](https://github.com/ballerina-platform/module-ballerina-http/pull/2202/commits/2aa962f18f1fcf259c65a9deae84606a1b2b4dad)

compiler-plugin/src/main/resources/rules.json
MohamedSabthar: There is an issue for report improvement. I have added this as a comment to track: https://github.com/ballerina-platform/static-code-analysis-tool/issues/17#issuecomment-2458920611

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpServiceDeclarationAnalyzer.java
MohamedSabthar: Refactored [2aa962f](https://github.com/ballerina-platform/module-ballerina-http/pull/2202/commits/2aa962f18f1fcf259c65a9deae84606a1b2b4dad)

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpServiceAnalyzer.java
TharmiganK: Ack 👍🏼

compiler-plugin/src/main/resources/rules.json
TharmiganK: Ack 👍🏼

TharmiganK (approved):
Implementation LGTM

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpRule.java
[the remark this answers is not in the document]
SasinduDilshara: Currently our plan is to not to include those details in the report. But those will include in our official documentation(Website).
Looping @MaryamZi

TharmiganK (approved):

PR #2151, opened 2024-09-18
[master] Use `toJson` method from data.jsondata module in http client

lnash94:
## Purpose
Fix: https://github.com/ballerina-platform/ballerina-library/issues/7012

Client data binding integration: https://github.com/ballerina-platform/ballerina-library/issues/6747
Service data binding integration: https://github.com/ballerina-platform/ballerina-library/issues/7011

Disable the below test because of data.json data module currently does not support the `table` types
and enable tests once this issue is done: https://github.com/ballerina-platform/ballerina-library/issues/7020

- testIntTableDatabinding()
- testIntTableOrXmlArrayDatabinding()
- testIntTableDatabindingByType()
- testStringTableDatabinding()
- testStringTableDatabindingByType()
- testRecordTableDatabinding()
- testRecordTableDatabindingByType()
- testByteArrTableDatabinding()
- testByteArrTableDatabindingByType()

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] Updated the spec
- [ ] Checked native-image compatibility
- [ ] Checked the impact on OpenAPI generation

Dependent PR : https://github.com/ballerina-platform/module-ballerina-data.jsondata/pull/35 once this is merged , test can be added with new data.jsondata timestamp

PR #2204, opened 2024-11-03
[master] Use `toJson` method from data.jsondata module in http client

lnash94:
## Purpose
Fix: https://github.com/ballerina-platform/ballerina-library/issues/7012

Client data binding integration: https://github.com/ballerina-platform/ballerina-library/issues/6747
Service data binding integration: https://github.com/ballerina-platform/ballerina-library/issues/7011

Disable the below test because the data.json data module currently does not support the `table` types
and enable tests once these issues are fixed.

1. Related issue : https://github.com/ballerina-platform/ballerina-library/issues/7020

- testIntTableDatabinding()
- testIntTableOrXmlArrayDatabinding()
- testIntTableDatabindingByType()
- testStringTableDatabinding()
- testStringTableDatabindingByType()
- testRecordTableDatabinding()
- testRecordTableDatabindingByType()
- testByteArrTableDatabinding()
- testByteArrTableDatabindingByType()

2. Related issue https://github.com/ballerina-platform/ballerina-library/issues/7326

- testResponseAnydataNegative() 
- testRequestAnydataNegative()

The issue for tracking the disabled tests: https://github.com/ballerina-platform/ballerina-library/issues/7021

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] Updated the spec

Dependent PR : https://github.com/ballerina-platform/module-ballerina-data.jsondata/pull/35 once this is merged , test can be added with new data.jsondata timestamp

Closed PR : https://github.com/ballerina-platform/module-ballerina-http/pull/2151

### DIFF
--- a/ballerina-tests/http-advanced-tests/Dependencies.toml
+++ b/ballerina-tests/http-advanced-tests/Dependencies.toml
@@ -5,7 +5,7 @@
 
 [ballerina]
 dependencies-toml-version = "2"
-distribution-version = "2201.10.0"
+distribution-version = "2201.11.0-20241008-112400-81975006"
 
 [[package]]
 org = "ballerina"
--- a/ballerina-tests/http-advanced-tests/Dependencies.toml
+++ b/ballerina-tests/http-advanced-tests/Dependencies.toml
@@ -56,6 +56,16 @@ modules = [
 
 [[package]]
 org = "ballerina"
+name = "data.jsondata"
+version = "0.3.0"
+scope = "testOnly"
+dependencies = [
+	{org = "ballerina", name = "jballerina.java"},
+	{org = "ballerina", name = "lang.object"}
+]
+
+[[package]]
+org = "ballerina"
 name = "file"
 version = "1.10.0"
 scope = "testOnly"
@@ -79,6 +89,7 @@ dependencies = [
 	{org = "ballerina", name = "cache"},
 	{org = "ballerina", name = "constraint"},
 	{org = "ballerina", name = "crypto"},
+	{org = "ballerina", name = "data.jsondata"},
 	{org = "ballerina", name = "file"},
 	{org = "ballerina", name = "io"},
 	{org = "ballerina", name = "jballerina.java"},
--- a/ballerina/http_client_payload_builder.bal
+++ b/ballerina/http_client_payload_builder.bal
@@ -162,7 +162,7 @@ isolated function jsonPayloadBuilder(Response response, TargetType targetType) r
 isolated function nonNilablejsonPayloadBuilder(Response response, typedesc<anydata> targetType)
         returns anydata|ClientError {
     json payload = check response.getJsonPayload();
-    var result = jsondata:parseAsType(payload, {enableConstraintValidation: false}, targetType);
+    var result = jsondata:parseAsType(payload, {enableConstraintValidation: false, allowDataProjection: false}, targetType);
     return result is error ? createPayloadBindingError(result) : result;
 }
 
@@ -170,7 +170,7 @@ isolated function nilablejsonPayloadBuilder(Response response, typedesc<anydata>
         returns anydata|ClientError {
     json|ClientError payload = response.getJsonPayload();
     if payload is json {
-        var result = jsondata:parseAsType(payload, {enableConstraintValidation: false}, targetType);
+        var result = jsondata:parseAsType(payload, {enableConstraintValidation: false, allowDataProjection: false}, targetType);
         return result is error ? createPayloadBindingError(result) : result;
     } else {
         return payload is NoContentError ? () : payload;
--- a/ballerina/http_client_payload_builder.bal
+++ b/ballerina/http_client_payload_builder.bal
@@ -162,7 +162,7 @@ isolated function jsonPayloadBuilder(Response response, TargetType targetType) r
 isolated function nonNilablejsonPayloadBuilder(Response response, typedesc<anydata> targetType)
         returns anydata|ClientError {
     json payload = check response.getJsonPayload();
-    var result = jsondata:parseAsType(payload, t= targetType);
+    var result = jsondata:parseAsType(payload, t = targetType);
     return result is error ? createPayloadBindingError(result) : result;
 }
 
@@ -170,7 +170,7 @@ isolated function nilablejsonPayloadBuilder(Response response, typedesc<anydata>
         returns anydata|ClientError {
     json|ClientError payload = response.getJsonPayload();
     if payload is json {
-        var result = jsondata:parseAsType(payload, t= targetType);
+        var result = jsondata:parseAsType(payload, t = targetType);
         return result is error ? createPayloadBindingError(result) : result;
     } else {
         return payload is NoContentError ? () : payload;
--- a/ballerina/http_client_payload_builder.bal
+++ b/ballerina/http_client_payload_builder.bal
@@ -162,7 +162,7 @@ isolated function jsonPayloadBuilder(Response response, TargetType targetType) r
 isolated function nonNilablejsonPayloadBuilder(Response response, typedesc<anydata> targetType)
         returns anydata|ClientError {
     json payload = check response.getJsonPayload();
-    var result = jsondata:parseAsType(payload, t = targetType);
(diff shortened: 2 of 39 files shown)


lnash94:
Attached the reported issues for jsondata module regarding this migration for tracking the fixes. https://github.com/ballerina-platform/ballerina-library/issues?q=+label%3Amodule%2Fdata.jsondata+author%3Alnash94+

ballerina/http_client_payload_builder.bal
```diff
@@ -161,15 +162,15 @@ isolated function jsonPayloadBuilder(Response response, TargetType targetType) r
 isolated function nonNilablejsonPayloadBuilder(Response response, typedesc<anydata> targetType)
         returns anydata|ClientError {
     json payload = check response.getJsonPayload();
-    var result = payload.fromJsonWithType(targetType);
+    var result = jsondata:parseAsType(payload, {enableConstraintValidation: false}, targetType);
```
TharmiganK, line 165: By default, `parseAsType` is enabled projection support. Since the `fromJsonWithType` does not have the projection support, we need to disabled it by default.
```suggestion
    var result = jsondata:parseAsType(payload, {enableConstraintValidation: false, allowDataProjection: false}, targetType);
```

TharmiganK, line 165: @SasinduDilshara How does constraint validation works in `jsondata` package? Does it get validated during the parsing stage or we validate after parsing the json?

ballerina/http_client_payload_builder.bal
```diff
@@ -161,15 +162,15 @@ isolated function jsonPayloadBuilder(Response response, TargetType targetType) r
...
     return result is error ? createPayloadBindingError(result) : result;
 }
 
 isolated function nilablejsonPayloadBuilder(Response response, typedesc<anydata> targetType)
         returns anydata|ClientError {
     json|ClientError payload = response.getJsonPayload();
     if payload is json {
-        var result = payload.fromJsonWithType(targetType);
+        var result = jsondata:parseAsType(payload, {enableConstraintValidation: false}, targetType);
```
TharmiganK, line 173: Same as before

ballerina/http_request.bal
```diff
@@ -414,17 +415,34 @@ public class Request {
 
     # Sets a `json` as the payload. If the content-type header is not set then this method set content-type
     # headers with the default content-type, which is `application/json`. Any existing content-type can be
-    # overridden by passing the content-type as an optional parameter.
+    # overridden by passing the content-type as an optional parameter. If the given payload is a record type 
+    # with the `@jsondata:Name` annotation, the `jsondata:toJson` function processes the name and populates
```
TharmiganK, line 419: Can't we just say that internally the records will be converted to json using the `jsondata:toJson` method?

native/build.gradle
```diff
@@ -42,7 +42,7 @@ dependencies {
     implementation group: 'io.ballerina.stdlib', name: 'io-native', version: "${stdlibIoVersion}"
     implementation group: 'io.ballerina.stdlib', name: 'mime-native', version: "${stdlibMimeVersion}"
     implementation group: 'io.ballerina.stdlib', name: 'constraint-native', version: "${stdlibConstraintVersion}"
-
+    implementation group: 'io.ballerina.lib', name: 'data.jsondata-native', version: "${stdlibDataJsonDataVersion}"
```
TharmiganK, line 45:
```suggestion
    implementation group: 'io.ballerina.lib', name: 'data.jsondata-native', version: "${stdlibDataJsonDataVersion}"
    
```
Just separate the ballerina related dependencies from the external java ones

native/src/main/java/io/ballerina/stdlib/http/api/service/signature/converter/JsonToRecordConverter.java
```diff
@@ -60,7 +67,14 @@ private static Object getRecordEntity(BObject entity, Type entityBodyType) {
      */
     private static Object getRecord(Type entityBodyType, Object bJson) {
         try {
-            return ValueUtils.convert(bJson, entityBodyType);
+            Map<String, Object> valueMap = new HashMap<>();
+            Boolean bool = Boolean.FALSE;
+            valueMap.put("enableConstraintValidation", bool);
```
TharmiganK, line 72: Define constants for this field names: `enableConstraintValidation` and `Options`

TharmiganK, line 72: We have to disable the projection as well

ballerina/http_client_payload_builder.bal
TharmiganK, line 165: Lets add a test case to validate the projection. For example if we return a json which has fields `a`, `b`, and `c` and try to data-bind it to a **closed** record with only `a` and `b`, then it should return an error since the projection is disabled by default

SasinduDilshara: It validate after parsing the json.

It will first parse the json and construct the record value.
Then it will validate that value.

ballerina-tests/http-advanced-tests/Dependencies.toml
```diff
@@ -5,7 +5,7 @@
 
 [ballerina]
 dependencies-toml-version = "2"
-distribution-version = "2201.10.0"
+distribution-version = "2201.11.0-20241008-112400-81975006"
```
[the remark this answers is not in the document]
lnash94: Yeah, because one of this fix is related to the lang https://github.com/ballerina-platform/ballerina-library/issues/7024, therefore we have to update the lang as well

ballerina/http_client_payload_builder.bal
lnash94: Updated

ballerina/http_request.bal
lnash94: Please check updated there 
271e798

ballerina/http_request.bal
```diff
@@ -414,17 +415,33 @@ public class Request {
...
+    # overridden by passing the content-type as an optional parameter. If the given payload is a record type with 
+    # the `@jsondata:Name` annotation, the `jsondata:toJson` function internally converts the record to JSON
     #
     # + payload - The `json` payload
     # + contentType - The content type of the payload. This is an optional parameter.
     #                 The `application/json` is the default value
     public isolated function setJsonPayload(json payload, string? contentType = ()) {
         mime:Entity entity = self.getEntityWithoutBodyAndHeaders();
-        setJson(entity, payload, self.getContentType(), contentType);
+        setJson(entity, jsondata:toJson(payload), self.getContentType(), contentType);
```
TharmiganK, line 426: This `jsondata:toJson` can panic right?

native/src/main/java/io/ballerina/stdlib/http/api/HttpConstants.java
```diff
@@ -497,7 +497,13 @@ public final class HttpConstants {
     public static final BString SOCKET_CONFIG_SOCKET_REUSE = StringUtils.fromString("socketReuse");
     public static final BString SOCKET_CONFIG_KEEP_ALIVE = StringUtils.fromString("keepAlive");
 
+    //Service parseAsType constants
+    public static final String ENABLE_CONSTRAINT_VALIDATION = "enableConstraintValidation";
+    public static final String ALLOW_DATA_PROJECTION = "allowDataProjection";
+    public static final String PARSER_AS_TYPE_OPTIONS =  "Options";
+
     //Client Endpoint (CallerActions)
+
```
TharmiganK, line 506: Remove the additional new line
```suggestion
```

ballerina/http_request.bal
lnash94: According to the `toJson` method signature. it only returns `json` type there is no error return. Yeah, there can be panic happened, if the user has provided the value which will lead to some runtime error. Are we going to handle it here or keep this as it is based on the `toJson` function?

TharmiganK: In this specific scenario, I don't think it will panic since the input is also a json type. So it is not required to handle that panics

TharmiganK (approved):
LGTM

PR #2218, opened 2024-11-18
Migrate to Java 21

warunalakshitha:
## Purpose
$Subject

## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec


native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
```diff
@@ -173,98 +166,54 @@ public static Object options(Environment env, BObject client, BString path, Obje
...
-                        HttpUtil.createHttpError("client method invocation failed: " + bError.getErrorMessage(),
-                                                 HttpErrorType.CLIENT_ERROR, bError);
-                balFuture.complete(invocationError);
-            }
-        }, propertyMap, PredefinedTypes.TYPE_NULL, paramFeed);
-        return null;
-    }
-
-    private static Map<String, Object> getPropertiesToPropagate(Environment env) {
-        String[] keys = {CURRENT_TRANSACTION_CONTEXT_PROPERTY, KEY_OBSERVER_CONTEXT, SRC_HANDLER, MAIN_STRAND,
```
TharmiganK, line 253: With the new runtime APIs, don't we support adding properties to the strand which can be obtained `Environment.getStrandLocal` API? We have some logic based on these properties

gradle.properties
```diff
@@ -26,26 +26,32 @@ jacocoVersion=0.8.10
...
+stdlibTaskVersion=2.5.1-20241113-123500-f905281
+stdlibLogVersion=2.10.1-20241113-120000-4577868
+stdlibCryptoVersion=2.7.3-20241113-081400-d015a39
 
-stdlibAuthVersion=2.12.0
-stdlibDataJsonDataVersion = 0.3.0-20241105-101100-661d11f
-stdlibJwtVersion=2.13.0
-stdlibOAuth2Version=2.12.0
+stdlibFileVersion=1.10.1-20241113-151700-e1a2e38
+stdlibMimeVersion=2.10.2-20241113-154200-d953747
```
TharmiganK, line 45: Shall we update the latest timestamped version? There is a change in the API which used to write streams: https://github.com/ballerina-platform/module-ballerina-mime/pull/546/commits/50d8dca19127899acae2065297f61d34d866d89e

native/src/main/java/io/ballerina/stdlib/http/api/client/actions/HttpClientAction.java
HindujaB: Do we have any tests covering the case? The strand local was not read anywhere.

warunalakshitha: @TharmiganK  you can dd new properties to strandmetadata

warunalakshitha: Here with new runtime APIs this code run in same strand so we do not need to copy properties. runtime will only run concurrently if it came from non strand thread.

TharmiganK: There are few use cases here: https://github.com/search?q=repo%3Aballerina-platform%2Fmodule-ballerina-http%20%22getStrandLocal%22&type=code

I think we are using it for access logging but we do not have any tests for that

warunalakshitha: Still those tests will pass since this code is coming from extern function call, same strand will use so as properties.

HindujaB (approved):
LGTM

PR #2223, opened 2024-11-21
[2201.7.x] Handle bal future complete panic by logging the error

TharmiganK:
## Purpose

> $Subject

The Bal Future complete will throw an error when we try to complete the same future twice(this is an unexpected behaviour). In the current implementation, we do not handle that error, and in some scenarios, this is ending up in a panic, which cause the service to crash. This PR is to catch that exception and add a log which can resolve the panic behaviour and will provide more context on where this error occur(this can be used to trace the unexpected completion).

## Examples

N/A

## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec


TharmiganK:
Closing this since this should be handled in the lang side. Related issue: https://github.com/ballerina-platform/ballerina-lang/issues/43654

PR #2427, opened 2025-07-04
Fix URL dispatching issues with special characters

TharmiganK:
## Purpose

Fixes: https://github.com/ballerina-platform/ballerina-library/issues/4154

Currently, Ballerina resource functions can be defined with special characters. To dispatch to that particular resource function the URL should encode all the special characters. But the URL specification specifies certain characters as reserved which can be present in the URL without explicit encoding. The actual reason to have these reserved characters is to use them in a special context (special delimiters) in the application layer. Even though, there is no special handling still we could have them without encoding. Since Ballerina always expects the special characters to be encoded there is no way for a request with allowed reserved character to dispatch to the service.

This PR makes the dispatching in a more relaxed way. When constructing the resource path segments from the signature we only encode the that should be encoded (skipping the reserved characters by default). When dispatching we compare the raw URL against this path segments. Additionally if we found an encoded character in the raw path and a special character in the path segment, then we decode the encoded character and compare it with the path segment.

## Examples

```ballerina
import ballerina/http;

service /api on new http:Listener(9090) {

    resource function get \$path() returns string => "Hello, World!";
}
```

Both the following URLs will be dispatched to the same resource:
```console
$ curl -v 'http://localhost:9090/api/$path'
```
```console
$ curl -v 'http://localhost:9090/api/%24path'
```

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] ~Updated the spec~


TharmiganK:
Making this draft since websocket tests are failing with this change: https://github.com/ballerina-platform/module-ballerina-http/actions/runs/16140016880/job/45545093395

ThisaruGuruge (approved):
LGTM

PR #2501, opened 2025-09-18
Implement static code analysis rule to detect unsecure HTTP redirections

nureka-rodrigo:
This includes the addition of the `HttpRedirectAnalyzer` class, which checks for potential vulnerabilities in function signatures that return temporary redirects. The rule aims to prevent open forging attacks by ensuring that redirect locations are not user-controlled.

## Purpose

Related to: https://github.com/ballerina-platform/ballerina-library/issues/8254

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] Updated the spec


compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/staticcodeanalyzer/HttpRedirectAnalyzer.java
```diff
@@ -0,0 +1,224 @@
...
+        analyzeImports(context);
+        FunctionSignatureNode functionSignature = (FunctionSignatureNode) context.node();
+        semanticModel = context.semanticModel();
+        if (functionSignature.returnTypeDesc().isEmpty()) {
+            return;
+        }
+        ReturnTypeDescriptorNode returnTypeDescriptor = functionSignature.returnTypeDesc().get();
+        if (!(returnTypeDescriptor.type() instanceof QualifiedNameReferenceNode qualifiedNameReference)
+                || !httpPrefixes.contains(qualifiedNameReference.modulePrefix().text().trim())
+                || !qualifiedNameReference.identifier().text().equals(TEMPORARY_REDIRECT)) {
```
ThisaruGuruge: This condition is hard to read. Let's try to break it down (Maybe a separate function `isTemporaryRedirect`?). 
I still advice on using Semantic API instead of string comparison in these cases.

I am not sure whether this works if the import is something like this:

```ballerina
import ballerina/http as h;
```

The current approach seems to _add_ HTTP prefixes to a map and then while validating, checks whether it qualifies _one of_ the prefixes. But this might be problematic. Consider the following scenario:


- **File `a.bal`**:

	```ballerina
	import ballerina/http as h;
	import myorg/http;
	```

- **File `b.bal`**:

    ```ballerina
    import ballerina/http;
    import myorg/http as h;
    ```

My point is when we compare strings, there are a lot of edge cases to consider and it would be hard to cover all the cases. But using the Semantic API, this wouldn't be a problem since we get the package `org` and `name` from the API itself. Therefore, we don't have to consider the aliases, we just have to check the original package name and the organization.

nureka-rodrigo: Fix: cce149f52b06ffb232c17c3d62faf55ccea9e959

Created a new helper method, `isTemporaryRedirect` which uses semantic model API to do the string comparison.

nureka-rodrigo: @ThisaruGuruge The current `analyzeImports` method analyzes both the `org` name and the `module` name. It also analyzes any custom prefix used for the module name. So, the first case you mentioned (`import ballerina/http as h`) is handled by the current implementation. But, the second case (`import myorg/http`) is not handled, since the implementation explicitly checks for the `ballerina` as the `org` name. This decision did not come from me but was adapted from previous rule implementations (for example, [see here](https://github.com/ballerina-platform/module-ballerina-log/blob/master/compiler-plugin/src/main/java/io/ballerina/stdlib/log/compiler/staticcodeanalyzer/LogStatementAnalyzer.java#L86)). If this logic is incorrect I could refactor the implementation to remove explicit `org` name check.

nureka-rodrigo: Also I agree with your point: _"when we compare strings, there are a lot of edge cases to consider"_.  Just to clarify, does using `text()` from the Ballerina compiler API instead of `toString()` from the standard Java API mitigate those issues? I was told to use `text()` whenever possible instead of `toString()`, and I assume this is the reason. So, if using `text()` does actually mitigate these issues, then using the semantic API for string comparison is not necessary, right?

PR #2511, opened 2025-09-26
Fix hanging issue when load testing with relatively larger payload

TharmiganK:
## Purpose

> $Subject

Fixes: https://github.com/ballerina-platform/ballerina-library/issues/8271

Load test results can be found in this comment: https://github.com/ballerina-platform/ballerina-library/issues/8271#issuecomment-3336625900

## Examples

N/A

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [ ] ~Added tests~
- [ ] ~Updated the spec~

### DIFF
--- a/native/src/main/java/io/ballerina/stdlib/http/transport/message/HttpCarbonMessage.java
+++ b/native/src/main/java/io/ballerina/stdlib/http/transport/message/HttpCarbonMessage.java
@@ -47,7 +47,6 @@ import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * HTTP based representation for HttpCarbonMessage.
@@ -84,8 +83,6 @@ public class HttpCarbonMessage {
     private Integer contentSize = 0;
     private boolean contentReleased = false;
 
-    private final ReentrantLock messageLock = new ReentrantLock();
-
     public HttpCarbonMessage(HttpMessage httpMessage, Listener contentListener) {
         this.httpMessage = httpMessage;
         setBlockingEntityCollector(new BlockingEntityCollector(Constants.ENDPOINT_TIMEOUT));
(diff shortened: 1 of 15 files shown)


native/src/main/java/io/ballerina/stdlib/http/transport/message/HttpCarbonMessage.java
```diff
@@ -694,14 +666,9 @@ public boolean isLastHttpContentArrived() {
...
      */
-    public void notifyContentFailure(Exception exception) {
-        messageLock.lock();
-        try {
-            if (fullHttpMessageFuture != null) {
-                fullHttpMessageFuture.notifyFailure(exception);
-            }
-        } finally {
-            messageLock.unlock();
+    public synchronized void notifyContentFailure(Exception exception) {
```
ThisaruGuruge, line 669: Does introducing synchronized methods have any performance hit?

native/src/main/java/io/ballerina/stdlib/http/transport/message/MessageFuture.java
```diff
@@ -33,15 +31,13 @@ public class MessageFuture {
...
-    private final ReentrantLock futureLock = new ReentrantLock();
 
     public MessageFuture(HttpCarbonMessage httpCarbonMessage) {
         this.httpCarbonMessage = httpCarbonMessage;
     }
 
     public void setMessageListener(MessageListener messageListener) {
-        futureLock.lock();
-        try {
+        synchronized (httpCarbonMessage) {
```
ThisaruGuruge, line 40: Ditto

native/src/main/java/io/ballerina/stdlib/http/transport/message/HttpCarbonMessage.java
TharmiganK: We are not introducing but reverting it back. Initially it was a synchronized block/method but we change it using the reentrant lock when migrating to Java 21, since there is a possibility of [pinning when it is getting executed in a virtual thread](https://medium.com/@minadev/java-virtual-threads-understanding-jdk-21-limitations-before-the-jdk-25-lts-release-7b3754228e44). But it causes unexpected locking when we are consuming larger payload which impacted the performance.

Actually, these code blocks are executed by netty threads and they are platform threads and not virtual threads (netty thread model is not loom friendly yet), so there is no issue with the synchronized blocks/methods here.

I tried two fixes:
- Adding proper locks to fix the issue
- Reverting the locks and use the synchronized blocks/methods as we were using until update 10

Both fixed the hanging(until times out) issue but the second fix gave significant performance improvement in load tests with larger payloads. Load test results can be found in this comment: https://github.com/ballerina-platform/ballerina-library/issues/8271#issuecomment-3336625900

cc: @warunalakshitha

native/src/main/java/io/ballerina/stdlib/http/transport/message/MessageFuture.java
TharmiganK: Addressed via https://github.com/ballerina-platform/module-ballerina-http/pull/2511#discussion_r2381443369

PR #2605, opened 2026-03-05
Add Capability to Generate and Export Endpoint Details File Along with OpenAPI Specification

niduniDK:
## Purpose
This PR introduces functionality to generate and export endpoint details into a separate file (`endpoint.yaml`) for each endpoint, alongside the corresponding OpenAPI specification.

The primary goal of this enhancement is to enable standardized service metadata extraction and export across different protocol types. This approach supports a consistent mechanism for capturing endpoint-level metadata, which is particularly useful for cloud-based services such as Devant. (https://github.com/wso2-enterprise/integration-engineering/issues/1060). 

## Examples
The functionality can be enabled using the build flag `--export-endpoints`. Running the build with this flag generates `endpoint.yaml` files for each endpoint alongside the corresponding OpenAPI specifications. A sample output folder structure is shown below.

```
target/
└── artifact/
    ├── health_endpoint.yaml
    ├── health_openapi.yaml
    ├── probes_endpoint.yaml
    └── probes_openapi.yaml
├── bin/
├── cache/
├── resources/
├── build

```

## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec

## Overview
Adds an opt-in build-time feature that generates and exports standardized endpoint metadata files (endpoint.yaml) alongside OpenAPI specifications for HTTP services. When the --export-endpoints build flag is enabled, the compiler plugin emits per-endpoint YAML artifacts into target/artifact next to OpenAPI YAML to support consistent service metadata extraction for tooling and cloud integrations.

## Key changes
- ServiceArtifactsExtractor: new syntax-node analysis task that coordinates conditional artifact generation during compilation, integrates with OpenAPI generation, respects the --export-endpoints flag, and includes a compatibility guard for older toolchains.
- EndpointYamlGenerator: computes endpoint metadata (port, basePath, type, schemaPath) from service declarations and OpenAPI server entries and writes YAML artifacts.
- FileNameGeneratorUtil: centralizes normalized, deduplicated filename generation for endpoint and OpenAPI outputs.
- Data models: new Endpoint and EndpointWrapper classes to represent endpoint metadata for serialization.
- OpenAPISpecGenerator: refactored constructFileName to public static for reuse.
- Tests: new ServiceArtifactsExtractionTest with golden-file validations covering artifact generation, filenames, and content, plus focused in-process checks for service node extraction and diagnostics.
- Build/module updates: added Jackson YAML/databind and Ballerina runtime compile-time dependencies, updated module-info, bumped package/module versions and test artifact paths, and updated gradle/manifest metadata.

## Behavior and error handling
- Exports run only when --export-endpoints is enabled; test sources are skipped and exports are suppressed when compilation errors are present.
- IO/write failures and mapper errors are converted to diagnostics reported during analysis.
- A compatibility guard handles missing build-option APIs on older toolchains to avoid hard failures.

## Impact
- Introduces new public Java classes and methods in the HTTP compiler plugin and adds compile-time dependencies for YAML serialization.
- No runtime behavior changes to services unless the build flag is used; enabling it produces additional artifact files alongside existing OpenAPI outputs.
- Review discussion noted compatibility concerns about adding build-option related APIs in patch releases and suggested consolidating shared artifact export utilities at the language level in future work.
- PR checklist items referenced in the description remain unchecked.


compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/Endpoint.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2024, WSO2 LLC. (http://www.wso2.com).
```
Bhashinee, line 2:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
```

gradle.properties
```diff
@@ -1,7 +1,7 @@
 org.gradle.caching=true
 group=io.ballerina.stdlib
 version=2.15.5-SNAPSHOT
```
TharmiganK: Lets bump the version to the next minor version since the feature is going to be released with update 14

gradle.properties
```diff
@@ -1,7 +1,7 @@
...
-ballerinaLangVersion=2201.13.0
+ballerinaLangVersion=2201.14.0-SNAPSHOT
```
TharmiganK: Bump the distribution version in here as well: https://github.com/ballerina-platform/module-ballerina-http/blob/5bc9ab221527cb9862cfa537251fe5583d3ddb06/build-config/resources/Ballerina.toml#L10

ballerina/Dependencies.toml
```diff
@@ -50,7 +50,7 @@ modules = [
 [[package]]
 org = "ballerina"
 name = "crypto"
-version = "2.9.3"
+version = "2.10.1"
```
TharmiganK: We need to revert this auto version bump from the local packages. We need to stick with the versions in the gradle.properties file

Check the associated auto commit and undo the unmatched version bumps

niduniDK:
> If I understand correctly, this endpoint and endpoint.yaml concept is a new concept in Ballerina language right? If so, I expect the necessary interfaces and utility functions should be provided by Ballerina language itself. This will make sure the template is defined and maintained by language rather than specific library packages and also remove the duplicate code requires to be added in each protocol library packages. If there is a requirement for libraries to have specific fields in the file, then also the base structure should remain in lang and it should be extensible in other libraries. Ideally what I expect is when the build option is enabled I need to extract the necessary details from the library's compiler plugin and call the lang functions to write or process the extracted information (structure and writing part is purely isolated to lang level)

Since this could not be added to 2201.13.2 release, it was decided to include this in each standard library level and it will be added to ballerina language later on.

gradle.properties
niduniDK: Bumped the version

TharmiganK:
> Since this could not be added to 2201.13.2 release, it was decided to include this in each standard library level and it will be added to ballerina language later on.

@niduniDK I believe we need a new API to check the build option for generating the service artifacts right? In that case a new API addition cannot be added to a patch version of lang, since the new http should be compatible with all the update 13 patches

@Bhashinee / @azinneera FYA

build-config/resources/Ballerina.toml
```diff
@@ -7,7 +7,7 @@ keywords = ["http", "network", "service", "listener", "client"]
 repository = "https://github.com/ballerina-platform/module-ballerina-http"
 icon = "icon.png"
 license = ["Apache-2.0"]
-distribution = "2201.13.0"
+distribution = "2201.14.0-SNAPSHOT"
```
TharmiganK:
```suggestion
distribution = "2201.13.3"
```

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/Endpoint.java
```diff
@@ -0,0 +1,49 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.ballerina.stdlib.http.compiler.endpointyaml.generator;
+
+public class Endpoint {
```
TharmiganK, line 24: Add proper documentations. Check other classes as well

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/EndpointYamlGenerator.java
```diff
@@ -0,0 +1,163 @@
...
+    private String schemaExtension = "";
+    private int portVal = 0;
+
+    private String type;
+    private final Server server;
+
+    public EndpointYamlGenerator(ServiceDeclarationNode node, SyntaxNodeAnalysisContext context, Server server) {
+        this.node = node;
+        this.context = context;
+        this.server = server != null ? new Server()
```
TharmiganK, line 69: Why do we copy the server information?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/EndpointYamlGenerator.java
```diff
@@ -0,0 +1,163 @@
...
+                .description(server.getDescription())
+                .variables(server.getVariables())
+                .extensions(server.getExtensions()) : null;
+
+        FileNameGeneratorUtil fileNameGeneratorUtil = new FileNameGeneratorUtil(context, this.schemaExtension);
+        this.schemaFileName = fileNameGeneratorUtil.getFileName();
+    }
+
+    public Endpoint getEndpoint() {
+        ServerVariables vars = server.getVariables();
```
TharmiganK, line 84: server can be null right?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/EndpointYamlGenerator.java
```diff
@@ -0,0 +1,163 @@
...
+        Path outPath = resolveOutputPath();
+        String fileName = buildEndpointFileName(outPath);
+        Path path = Paths.get(TARGET, ARTIFACT, fileName + YAML_EXTENSION).toAbsolutePath();
+        writeYaml(path, new EndpointWrapper(ep));
+    }
+
+    private String getBasePath() {
+        StringBuilder serviceBasePath = new StringBuilder();
+        NodeList<Node> resourcePathNode = node.absoluteResourcePath();
+        for (Node identifierNode : resourcePathNode) {
```
TharmiganK, line 112: Resource path can be a string literal or an absolute path. Are we handling both?
- `service "/api" on ...`
- `service /api on ...`

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/EndpointYamlGenerator.java
```diff
@@ -0,0 +1,163 @@
...
+            serviceBasePath.append(identifierNode.toString().replace("\"", "").trim());
+        }
+        return serviceBasePath.toString();
+    }
+
+    private Path resolveOutputPath() {
+        Package currentPackage = this.context.currentPackage();
```
TharmiganK, line 119: Verify whether this works for single files and work spaces

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/EndpointYamlGenerator.java
```diff
@@ -0,0 +1,163 @@
...
+            Files.createDirectories(Paths.get(String.valueOf(outPath), ARTIFACT));
+        } catch (IOException e) {
+            outStream.println(e);
+        }
+        return outPath;
+    }
+
+    private String buildEndpointFileName(Path outPath) {
+        String base;
+        if (REST.equals(this.type)) {
```
TharmiganK: Type is always REST for HTTP right?
Do not complicate the logic by having generic components. As we agreed in the meeting, ideally these should be moved to ballerina lang. So, having specific logic in the libraries should be enough

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/EndpointYamlGenerator.java
```diff
@@ -0,0 +1,163 @@
...
+        YAMLFactory yamlFactory = YAMLFactory.builder()
+                .disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER)
+                .build();
+        ObjectMapper mapper = new ObjectMapper(yamlFactory);
+        mapper.findAndRegisterModules();
+
+        try (Writer writer = Files.newBufferedWriter(path)) {
+            mapper.writeValue(writer, wrapper);
+        } catch (IOException e) {
+            outStream.println(e);
```
TharmiganK: Printing the error as it is?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/FileNameGeneratorUtil.java
```diff
@@ -0,0 +1,166 @@
...
+import static io.ballerina.openapi.service.mapper.Constants.YAML_EXTENSION;
+import static io.ballerina.openapi.service.mapper.utils.MapperCommonUtils.getNormalizedFileName;
+
+public class FileNameGeneratorUtil {
+
+    private static final String SLASH = "/";
+    private static final String UNDERSCORE = "_";
+    private static final String HYPHEN = "-";
+    private final String extension;
+    private final Map<Integer, String> services = new HashMap<>();
```
TharmiganK, line 51: Why do we need to do this pre-processing and load everything to a map?
Can't we generate the file name from base path when we process each service?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/OpenAPISpecGenerator.java
```diff
@@ -151,24 +154,6 @@ public void perform(SyntaxNodeAnalysisContext context) {
...
     }
 
-    /**
-     * This util function is to construct the generated file name.
-     *
-     * @param syntaxTree syntax tree for check the multiple services
-     * @param services   service map for maintain the file name with updated name
-     * @param serviceSymbol symbol for taking the hash code of services
-     */
-    private String constructFileName(SyntaxTree syntaxTree, Map<Integer, String> services, Symbol serviceSymbol) {
```
TharmiganK, line 161: Removing this logic and using a different approach to construct the file name is a breaking change
Lets keep this as it is

gradle.properties
```diff
@@ -1,7 +1,7 @@
 org.gradle.caching=true
 group=io.ballerina.stdlib
-version=2.16.1-SNAPSHOT
-ballerinaLangVersion=2201.13.0
+version=2.17.0-SNAPSHOT
+ballerinaLangVersion=2201.13.3-SNAPSHOT
```
TharmiganK: Get a timestamped version from lang time with the relevant fixes and update here. This SNAPSHOT version is not resolved

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/Endpoint.java
niduniDK: Ack. Added in [05422ae](https://github.com/ballerina-platform/module-ballerina-http/pull/2605/commits/05422aee0b2d69248515f7db45f7f98de19496e1)

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/EndpointYamlGenerator.java
niduniDK: When assigning the server directly, spotbugs task failed suggesting an exposure of internal representation by storing the server directly.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/EndpointYamlGenerator.java
niduniDK: Yes. Added a check in [05422ae](https://github.com/ballerina-platform/module-ballerina-http/pull/2605/commits/05422aee0b2d69248515f7db45f7f98de19496e1)

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/EndpointYamlGenerator.java
niduniDK: Both are handled in the same way

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/EndpointYamlGenerator.java
niduniDK: Tested for single files and multiple files

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/EndpointYamlGenerator.java
niduniDK: Updated in [05422ae](https://github.com/ballerina-platform/module-ballerina-http/pull/2605/commits/05422aee0b2d69248515f7db45f7f98de19496e1)

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/FileNameGeneratorUtil.java
niduniDK: This is added to handle the case where the file contains different services with same base path but listening to different ports.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/OpenAPISpecGenerator.java
niduniDK: Added it back. But change the visibility of the method to `public` as it is used in `ServiceArtifactGenerator.java` as well.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/FileNameGeneratorUtil.java
TharmiganK: Did not get this? My question is why do we need to store this map? IIUC the requirement is to extract the base path for a particular service on demand right?

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/OpenAPISpecGenerator.java
TharmiganK: Shall we add it back in the same location, so that the diff is shown correctly

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/FileNameGeneratorUtil.java
niduniDK: Yes, but the thing is if there are multiple services with same base path (with different ports), the endpoint.yaml will be overwritten. Since the filenames are same for all those.
This map is used to check whether the base path has already been detected, if so, the respective hashcode is added to the end of the base path while constructing the file name.
This map holds the service name (will be the parsed basepath, when it is not duplicated) and the respective hashcode of the service symbol.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/OpenAPISpecGenerator.java
niduniDK: Updated.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/FileNameGeneratorUtil.java
TharmiganK: If a code modifier get engaged, then it can change the hash code of the service symbol if there is a change in the service code. Are we highly depending on the service hash code to resolve the base path?

niduniDK: The hashcode is used to avoid file overwriting, when there are two services with same base path. This is about resolving file name. Basepath would be same for both services.

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/endpointyaml/generator/ServiceArtifactsExtractor.java
```diff
@@ -0,0 +1,238 @@
...
+        for (Server server: servers) {
+            EndpointYamlGenerator endpointYamlGeneratorHttp =
+                    new EndpointYamlGenerator(serviceNode, context, server);
+            try {
+                endpointYamlGeneratorHttp.writeEndpointYaml();
+            } catch (IOException e) {
+                diagnostics.add(getDiagnostics(
+                     new ExceptionDiagnostic(DiagnosticMessages.OAS_CONVERTOR_108, e.toString())));
+            }
+        }
```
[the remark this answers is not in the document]
niduniDK: This is tested and the endpoint.yaml filename is differentiated using the combination of basepath, servicenode's hash and and source file name.

TharmiganK (dismissed):
Code changes LGTM

compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/ServiceArtifactsExtractionTest.java
```diff
@@ -0,0 +1,899 @@
...
+        Assert.assertEquals(yamlFiles.stream()
+                .map(this::safeFileName)
+                .filter(fileName -> fileName.contains("_endpoint"))
+                .count(), 5L, "Expected 5 endpoint artifact files");
+
+        deleteDirectories(projectDirPath);
+    }
+
+    @Test
+    public void testEndpointYamlContentForSimpleService() throws Exception {
```
Bhashinee: Rather than having two test cases to check if files are there(testServiceArtifactGenerationWithSimpleService) and the content validation, let's have a single test case which validates both

compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/ServiceArtifactsExtractionTest.java
```diff
@@ -0,0 +1,899 @@
...
+            Assert.assertTrue(Files.exists(artifactDir.resolve("service_endpoint.yaml")),
+                    "Endpoint artifact file should be generated");
+        } finally {
+            deleteDirectories(projectDirPath);
+        }
+    }
+
+    @Test
+    public void testServiceArtifactGenerationWithoutFlag() throws Exception {
+        Path projectDirPath = RESOURCE_DIRECTORY.resolve("sample_package_20");
```
Bhashinee, line 111: Better to follow the try finally approach in other tests as you've done in `testServiceArtifactGenerationWithSimpleService`

compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/ServiceArtifactsExtractionTest.java
```diff
@@ -0,0 +1,899 @@
...
+                    "getDiagnostics must preserve the diagnostic severity");
+
+        } finally {
+            deleteDirectories(projectDirPath);
+        }
+    }
+
+    @Test
+    public void testFileNameGeneratorForNonServiceNode() {
+    Path projectDirPath = RESOURCE_DIRECTORY.resolve("sample_package_20");
```
Bhashinee: can you check the formatting? Please format the entire file

compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/ServiceArtifactsExtractionTest.java
niduniDK: Ack. Update in [78b6acf](https://github.com/ballerina-platform/module-ballerina-http/pull/2605/commits/78b6acf26b0ff6dd3ef12bff06233969e9285a07).

compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/ServiceArtifactsExtractionTest.java
niduniDK: Ack. Update in [78b6acf](https://github.com/ballerina-platform/module-ballerina-http/pull/2605/commits/78b6acf26b0ff6dd3ef12bff06233969e9285a07).

compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/ServiceArtifactsExtractionTest.java
niduniDK: Ack. Update in [78b6acf](https://github.com/ballerina-platform/module-ballerina-http/pull/2605/commits/78b6acf26b0ff6dd3ef12bff06233969e9285a07).

pcnfernando (dismissed):

lnash94:
Hi @niduniDK,

Here, I noticed that a separate endpoint file is generated for each service. Would it be possible to have a single `endpoints.yaml` file to maintain all endpoint details in one place as a list? I think this could make the structure easier to manage and, in the future, would allow us to directly point to a single file for endpoint configurations.

```yml
endpoints:
  - name: <endpoint_name>
    port: 8080
    basePath: "/userservice"
    type: "REST"
    schemaPath: "service_userservice_openapi.yaml"
```

niduniDK:
> Hi @niduniDK,
> 
> Here, I noticed that a separate endpoint file is generated for each service. Would it be possible to have a single `endpoints.yaml` file to maintain all endpoint details in one place as a list? I think this could make the structure easier to manage and, in the future, would allow us to directly point to a single file for endpoint configurations.
> 
> ```yaml
> endpoints:
>   - name: <endpoint_name>
>     port: 8080
>     basePath: "/userservice"
>     type: "REST"
>     schemaPath: "service_userservice_openapi.yaml"
> ```

Hi @lnash94 , 
Sure, we will update the logic to generate a single endpoints.yaml as a separate fix.

pcnfernando:
> > Hi @niduniDK,
> > Here, I noticed that a separate endpoint file is generated for each service. Would it be possible to have a single `endpoints.yaml` file to maintain all endpoint details in one place as a list? I think this could make the structure easier to manage and, in the future, would allow us to directly point to a single file for endpoint configurations.
> > ```yaml
> > endpoints:
> >   - name: <endpoint_name>
> >     port: 8080
> >     basePath: "/userservice"
> >     type: "REST"
> >     schemaPath: "service_userservice_openapi.yaml"
> > ```
> 
> Hi @lnash94 , Sure, we will update the logic to generate a single endpoints.yaml as a separate fix.

@niduniDK could you create an issue and point here itself so that we won't loose track of the improvement

niduniDK:
> > > Hi @niduniDK,
> > > Here, I noticed that a separate endpoint file is generated for each service. Would it be possible to have a single `endpoints.yaml` file to maintain all endpoint details in one place as a list? I think this could make the structure easier to manage and, in the future, would allow us to directly point to a single file for endpoint configurations.
> > > ```yaml
> > > endpoints:
> > >   - name: <endpoint_name>
> > >     port: 8080
> > >     basePath: "/userservice"
> > >     type: "REST"
> > >     schemaPath: "service_userservice_openapi.yaml"
> > > ```
> > 
> > 
> > Hi @lnash94 , Sure, we will update the logic to generate a single endpoints.yaml as a separate fix.
> 
> @niduniDK could you create an issue and point here itself so that we won't loose track of the improvement

Issue for the suggested improvement:  https://github.com/wso2/product-integrator/issues/1427

pcnfernando (approved):
LGTM

TharmiganK (approved):

PR #2611, opened 2026-03-25
Fix scope validation to require all expected scopes instead of partial match

VellummyilumVinoth:
## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec

This pull request tightens scope validation so that all expected scopes must be present in the actual scopes (no partial matches). The change makes authorization checks stricter by requiring complete scope compliance.

Key changes
- Core logic
  - Modified matchScopes in ballerina/auth_utils.bal to require that every expected scope is present in the actual scopes. The function now returns false immediately if any expected scope is missing and emits debug logging on failures; existing behavior for expectedScopes as a single string is preserved.

- Tests
  - Added four isolated tests in ballerina-tests/http-security-tests/tests/auth_listener_auth_handler_test.bal to verify authorization succeeds when all required scopes are present and fails when any expected scope is missing. Tests cover both file-based BasicAuth and JWT auth handlers.
  - Adjusted several existing security tests and test utilities: some test assertions and JWT test inputs were updated to align with the stricter scope semantics (e.g., JWT tokens and client custom claims updated).

- Version and dependency updates
  - Bumped package/test module versions from 2.16.0 → 2.16.1 across multiple test modules and updated native/test artifact paths to corresponding 2.16.1-SNAPSHOT jars.
  - Updated generated Dependencies.toml files to newer test/transitive dependency versions (including log and observe).
  - Two automated commits updated native JAR references.

Notes
- PR template fields for Purpose and Examples are empty. Checklist items for linked issue, changelog update, tests added, spec update, native-image compatibility check, and OpenAPI impact check are present but remain unchecked.

### DIFF
--- a/ballerina/auth_utils.bal
+++ b/ballerina/auth_utils.bal
@@ -95,7 +95,7 @@ isolated function extractAuthorizationHeader(Request|Headers|string data) return
     }
 }
 
-// Match the expectedScopes with actualScopes and return if there is a match.
+// Match the expectedScopes with actualScopes and return if expectedScopes is a subset of actualScopes.
 isolated function matchScopes(string|string[] actualScopes, string|string[] expectedScopes) returns boolean {
     if expectedScopes is string {
         if actualScopes is string {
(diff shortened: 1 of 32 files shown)


daneshk:
@VellummyilumVinoth This will be a breaking change.

ballerina/auth_utils.bal
```diff
@@ -110,18 +110,29 @@ isolated function matchScopes(string|string[] actualScopes, string|string[] expe
     } else {
         if actualScopes is string {
             foreach string expectedScope in expectedScopes {
```
TharmiganK, line 112: Can't we do like this:

```bal
            // A single-scope token can never satisfy multiple required scopes
            if expectedScopes.length() > 1 {
                log:printDebug("Failed to match the scopes. Expected '" + expectedScopes.toString() +
                               "', but found '" + actualScopes + "'");
                return false;
            }
            return actualScopes == expectedScopes[0];
```

VellummyilumVinoth: Ack We can do that

TharmiganK (changes requested):
This PR introduces a breaking change in the scope validation. Hence do not merge until approval

TharmiganK:
Closing this since it is introducing a breaking change

PR #2635, opened 2026-06-23
feat: support SOCKS proxy support for HTTP client

### LINKED ISSUE #8810: Add support for SOCKS protocol based proxies in the HTTP client
### Area

Area/Library

### Problem

Currently the HTTP client only support HTTP/HTTPS protocol based proxies and SOCKS protocol based proxies are not supported

### Proposed Solution

Add native support for SOCKS protocol based proxies and expose the configuration as below in the HTTP client configuration:

```ballerina
# Supported protocols of the proxy servers
enum ProxyProtocol {
    HTTP = "http",
    SOCKS4 = "socks4",
    SOCKS5 = "socks5"
}

# Proxy server configurations to be used with the HTTP client endpoint.
#
# + protocol - Protocol of the proxy server
# + host - Host name of the proxy server
# + port - Proxy server port
# + userName - Proxy server username
# + password - proxy server password
public type ProxyConfig record {|
    ProxyProtocol protocol = HTTP;
    string host = "";
    int port = 0;
    string userName = "";
    string password = "";
|};
```

### Alternatives

N/A

### Version

Ballerina SwanLake Update 13

ThisaruGuruge:
## Purpose

$Subject

Fixes: https://github.com/ballerina-platform/ballerina-library/issues/8810

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec

Added native SOCKS proxy support to the Ballerina HTTP client. This introduces new proxy configuration APIs (ProxyProtocol and an extended ProxyConfig with protocol selection) and updates the HTTP transport implementation to correctly route requests for HTTP, SOCKS4, and SOCKS5 proxies, including handling of proxy credentials (SOCKS4 password settings are ignored with a warning). The change also updates the HTTP client spec and documentation (including a new SOCKS proxy example) and adds/extends automated coverage using embedded and Docker-based SOCKS proxy servers, covering both end-to-end request behavior and SOCKS-based connection pooling. Supporting test/build infrastructure was updated to manage SOCKS proxy lifecycles and align dependency/version pins, along with an “Unreleased” changelog entry for the feature.


native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/sender/HttpClientChannelInitializer.java
```diff
@@ -176,19 +178,49 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
         ctx.close();
     }
 
-    // Use netty proxy handler only if scheme is https
+    // Configures the proxy handler in the pipeline based on the configured proxy protocol.
+    // For HTTP proxies, the netty proxy handler is used only if the scheme is https (sslConfig != null).
+    // For SOCKS4/SOCKS5 proxies, the handler is added regardless of sslConfig.
     private void configureProxyServer(ChannelPipeline clientPipeline) {
```
TharmiganK, line 184: Refactor this method to reduce its Cognitive Complexity[from SonarCloud]

native/src/main/java/io/ballerina/stdlib/http/api/HttpConstants.java
```diff
@@ -569,6 +569,10 @@ public final class HttpConstants {
     public static final BString PROXY_PORT = StringUtils.fromString("port");
     public static final BString PROXY_USERNAME = StringUtils.fromString("userName");
     public static final BString PROXY_PASSWORD = StringUtils.fromString("password");
+    public static final BString PROXY_PROTOCOL = StringUtils.fromString("protocol");
```
TharmiganK, line 572: There seems to be multiple constants already defined with the same value - `StringUtils.fromString("protocol")`

examples/socks-proxy/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## SOCKS4 vs SOCKS5
+
+| Capability                | SOCKS4 | SOCKS5 |
+|---------------------------|--------|--------|
+| Username/password auth    | No     | Yes    |
+| Remote (proxy-side) DNS   | No     | Yes    |
+
+- **Authentication:** SOCKS4 does not support password authentication. If a
+  `password` is set on a `SOCKS4` proxy configuration it is ignored. Use
```
TharmiganK: We are returning an error when password is passed for SOCKS4 right? Seems like there is some inconsistencies in the documentations

Also I prefer to print a warning and ignore password for SOCKS4. WDYT?

examples/socks-proxy/README.md
TharmiganK, line 1: Seems like the MD file text is already wrapped by using some length limit. I think we do not follow it as a practice in our documentations. We usually add content without defined text wrap

ballerina-tests/resources/socks/sockd.conf
```diff
@@ -0,0 +1,36 @@
+# Dante SOCKS server configuration for HTTP client SOCKS-proxy tests.
+#
+# Accepts SOCKS4 and SOCKS5 with no proxy authentication (socksmethod: none). Proxy
+# authentication (SOCKS5 username/password, SOCKS4 user id) and the SOCKS4 client-side-DNS
+# path are exercised by the native Java tests (Socks4/Socks5ProxyServerTestCase); this server
+# only needs to prove end-to-end SOCKS5 relaying for the Ballerina client.
+#
+# Listens on the published proxy port and relays to the backend Ballerina services running on
+# the host (reached via host.docker.internal — see compose.yml).
```
TharmiganK: It claims the proxy relays to backends "on the host (reached via host.docker.internal)", but it actually relays to the in-compose backend service. [Finding from local CodeRabbit and it suggests: "Listens on the published proxy port and relays to the in-compose backend service (resolved by service name via Docker DNS — see compose.yml)."]

ballerina-tests/build.gradle
```diff
@@ -300,6 +300,55 @@ task stopLdapServer() {
     }
 }
 
+task startSocksServer() {
+    doLast {
+        // The Dante image uses host networking, which is not supported on Windows Docker.
```
TharmiganK: We do not need this comment since we are not running any tests depends on docker in windows since Docker isn't reliably available on the Windows CI runners

Also local CodeRabbit review suggests that the compose file uses normal bridge networking with port publishing and not using host networking

native/src/main/java/io/ballerina/stdlib/http/api/HttpConstants.java
ThisaruGuruge: The two existing constants that share the same value serve different semantics:

- `SECURESOCKET_CONFIG_PROTOCOL` (line 472) accesses the `protocol` sub-record inside a `SecureSocket` config, which holds TLS settings like protocol name and versions — unrelated to proxying.
- `SERVICE_ENDPOINT_PROTOCOL_FIELD` (line 588) is defined but never referenced anywhere in the codebase (dead constant).

Neither name is appropriate to use for proxy config access. The codebase also already has this same pattern for other proxy fields — `PROXY_HOST`, `ENDPOINT_CONFIG_HOST`, `REMOTE_HOST_FIELD`, and `LOCAL_HOST_FIELD` all hold `"host"`, and each is kept separate for clarity. 

WDYT?

examples/socks-proxy/README.md
ThisaruGuruge: Updated the code to log a warning and continue.

native/src/main/java/io/ballerina/stdlib/http/api/HttpConstants.java
TharmiganK: Then lets keep it as it is

TharmiganK (approved):
LGTM

PR #2645, opened 2026-07-24
Fix security vulnerabilities with netty

TharmiganK:
## Purpose

> $Subject

Part of https://github.com/ballerina-platform/ballerina-library/issues/8924

Trivy Scan: https://github.com/ballerina-platform/module-ballerina-http/actions/runs/30070647893/job/89410605259

## Examples

N/A

## Checklist
- [x] Linked to an issue
- [ ] ~Updated the specification~
- [x] Updated the changelog
- [x] Added tests

## Summary

- Updated Netty to `4.1.136.Final` and Netty tcnative to `2.0.79.Final`.
- Synchronized Java 21 GraalVM native JAR references with the updated versions.
- Refreshed related Ballerina standard-library and Observe component dependencies.
- Regenerated dependency lock files across HTTP and integration tests.
- Added the update to the unreleased changelog.


ballerina/Dependencies.toml
```diff
@@ -50,7 +50,7 @@ modules = [
 [[package]]
 org = "ballerina"
 name = "crypto"
-version = "2.12.1"
+version = "2.9.3"
```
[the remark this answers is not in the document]
TharmiganK: This seems to be updated without any requirement. Ideally we should not bump the dependent library dependencies to a minor version unless the library actually uses those features. We should make sure we keep the minimum compatible minor version and patches we should always update

ballerina-tests/http-advanced-tests/Dependencies.toml
```diff
@@ -44,7 +44,7 @@ dependencies = [
 [[package]]
 org = "ballerina"
 name = "crypto"
-version = "2.12.1"
+version = "2.9.3"
```
ThisaruGuruge, line 47: Why this got downgraded?

ThisaruGuruge (approved):
LGTM, except for downgraded Crypto dependency. Seems like an issue with the build script?

ballerina-tests/http-advanced-tests/Dependencies.toml
TharmiganK: This was incorrectly updated in this PR: https://github.com/ballerina-platform/module-ballerina-http/pull/2605/changes#diff-a405ef45ae4a8f9db0a0b096036e0adb565a418d7a3924930daeed354ad971eeL47

The actual minimum compatible version is 2.9.1 - https://github.com/ballerina-platform/module-ballerina-http/blob/cf821305aab6a77de48826ace960b209d7c097f9/gradle.properties#L44 which is updated to the latest patch 2.9.3

Since we have not released a version with this incorrect change, now we could fix this rather than publishing http with crypto - 2.12.1

PR #2666, opened 2026-08-12
[2.16.x] fix(websocket): handle RFC 6455 simultaneous close correctly

ThisaruGuruge:
## Purpose

Fixes ballerina-platform/ballerina-library#9031.

`WebSocketInboundFrameHandler.notifyCloseMessage` required a peer's close frame to echo back the
exact status code this side sent when a close was already pending locally, failing the close with
an `IllegalStateException` otherwise. RFC 6455 §5.5.1 only says a responder "typically echoes" the
code (non-normative) and defines close completion as "after both sending and receiving a Close
message" — no code-equality requirement. A genuine simultaneous close (each side independently
sending its own close frame with its own code around the same time) is spec-valid and should
succeed; this code failed it instead.

Also fixes a related race in `DefaultWebSocketConnection.initiateConnectionClosure`: the close
promise was registered inside the outbound write's completion listener instead of before the write,
so a peer close frame arriving mid-write was misfiled as an unrelated unsolicited close, leaving the
promise to wait for an echo that would never come — until the default close timeout (60s) elapsed.
Under CPU contention this was observed stalling for the full 60s rather than failing fast.

## Examples

Found via `module-ballerina-graphql`'s server-side keep-alive tests, where a client's own graceful
`close()` (code 1000) can race the server's keep-alive-triggered forced close (code 4408). Before
this fix, that failed 100% of clean runs under a 2-core repro matching the GitHub Actions runner
profile — either immediately with `Expected status code 1000 but found 4408 in echoed close frame
from remote endpoint`, or, in some runs, stalling the full 60s close timeout. After this fix: 3/3
clean runs, no failures.

`testSendAndReceiveDifferentStatusCode` (native test suite) is updated to assert the RFC-compliant
behavior — the close now succeeds despite the code mismatch — instead of the old exception.

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [ ] Added tests (existing `testSendAndReceiveDifferentStatusCode` updated to match correct behavior; no new test added)
- [ ] Updated the spec


ThisaruGuruge:
Applied the same javadoc and flaky-test fixes that came out of review on #2667 (same PR, targeting a different base branch) — see that PR's review threads for the detailed rationale. 45ff2e44b.

TharmiganK (approved):

TharmiganK (approved):

PR #2667, opened 2026-08-12
[2.15.x] fix(websocket): handle RFC 6455 simultaneous close correctly

ThisaruGuruge:
## Purpose

Fixes ballerina-platform/ballerina-library#9031.

`WebSocketInboundFrameHandler.notifyCloseMessage` required a peer's close frame to echo back the
exact status code this side sent when a close was already pending locally, failing the close with
an `IllegalStateException` otherwise. RFC 6455 §5.5.1 only says a responder "typically echoes" the
code (non-normative) and defines close completion as "after both sending and receiving a Close
message" — no code-equality requirement. A genuine simultaneous close (each side independently
sending its own close frame with its own code around the same time) is spec-valid and should
succeed; this code failed it instead.

Also fixes a related race in `DefaultWebSocketConnection.initiateConnectionClosure`: the close
promise was registered inside the outbound write's completion listener instead of before the write,
so a peer close frame arriving mid-write was misfiled as an unrelated unsolicited close, leaving the
promise to wait for an echo that would never come — until the default close timeout (60s) elapsed.
Under CPU contention this was observed stalling for the full 60s rather than failing fast.

## Examples

Found via `module-ballerina-graphql`'s server-side keep-alive tests, where a client's own graceful
`close()` (code 1000) can race the server's keep-alive-triggered forced close (code 4408). Before
this fix, that failed 100% of clean runs under a 2-core repro matching the GitHub Actions runner
profile — either immediately with `Expected status code 1000 but found 4408 in echoed close frame
from remote endpoint`, or, in some runs, stalling the full 60s close timeout. After this fix: 3/3
clean runs, no failures.

`testSendAndReceiveDifferentStatusCode` (native test suite) is updated to assert the RFC-compliant
behavior — the close now succeeds despite the code mismatch — instead of the old exception.

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [ ] Added tests (existing `testSendAndReceiveDifferentStatusCode` updated to match correct behavior; no new test added)
- [ ] Updated the spec

## Summary

- Fixed WebSocket simultaneous close handling to support different status codes.
- Registered the close promise before sending the outbound close frame.
- Prevented close operations from waiting for the 60-second timeout.
- Updated native tests and the changelog.


native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/websocket/DefaultWebSocketConnection.java
```diff
@@ -186,11 +185,9 @@ public ChannelFuture initiateConnectionClosure() {
 
     private ChannelFuture initiateConnectionClosure(CloseWebSocketFrame closeWebSocketFrame) {
         handleCloseFrameSent();
-        closeInitiatedStatusCode = closeWebSocketFrame.statusCode();
-        closeInitiatedStatusCode = closeInitiatedStatusCode == -1 ? 1005 : closeInitiatedStatusCode;
         ChannelPromise closePromise = ctx.newPromise();
+        frameHandler.setClosePromise(closePromise);
```
TharmiganK, line 189: Registering the close promise *before* the write is the right fix for the simultaneous-close race, but it opens a new double-completion path that didn't exist when `setClosePromise` was called from inside the write listener.

Both completion sites are unconditional:
- here: `ctx.close().addListener(closeFuture -> closePromise.setFailure(cause))`
- `WebSocketInboundFrameHandler.notifyCloseMessage()`: `closePromise.setSuccess()`

Sequence that now breaks (peer sends close and then resets the connection):
1. app thread calls `initiateConnectionClosure()` → `closePromise` is visible to the frame handler immediately;
2. event loop reads the peer's close frame → `closePromise.setSuccess()`;
3. the pending `writeAndFlush` fails (connection reset) → listener runs → `closePromise.setFailure(cause)` on an already-completed promise → `IllegalStateException: complete already`, swallowed as a Netty listener warning.

The reverse order is worse: write fails first (`setFailure`), then an already-buffered close frame is read and `setSuccess()` throws straight out of `channelRead`, which lands in `exceptionCaught` and surfaces a spurious `onError` plus an extra close frame to the application.

Before this change step 2 could not precede step 3, because the promise was only published inside the write listener. Suggest switching both sites to the non-throwing variants (`closePromise.trySuccess()` / `closePromise.tryFailure(cause)`) so whichever event wins the race decides the outcome.

Separately: `WebSocketInboundFrameHandler.closePromise` is a plain non-volatile field. It used to be written on the event loop (inside the write listener) and read on the event loop; it is now written by whatever thread calls `initiateConnectionClosure()`, which is typically not the event loop. Marking the field `volatile` makes the publication explicit rather than relying on the incidental happens-before edge from the task submission in `writeAndFlush`.

native/src/test/java/io/ballerina/stdlib/http/transport/websocket/server/WebSocketServerFunctionalityTestCase.java
```diff
@@ -188,32 +188,20 @@ private WebSocketTestClient commandServerInitiatedClosure() throws URISyntaxExce
...
-        // after client echoing the close frame.
-        Assert.assertTrue(serverConnectorListener.getCloseFuture().isDone());
-        Assert.assertFalse(serverConnectorListener.getCloseFuture().isSuccess());
-        Assert.assertNotNull(cause);
-        Assert.assertTrue(cause instanceof IllegalStateException);
-        Assert.assertEquals(cause.getMessage(),
-                            String.format("Expected status code %d but found %d in echoed close frame from remote " +
-                                                  "endpoint",
-                                          closeFrame.statusCode(), closeFrame.statusCode() + 1));
+        Assert.assertTrue(serverConnectorListener.getCloseFuture().isSuccess());
```
TharmiganK, line 204: This test now only covers the status-code-check removal — with the mismatched code it is otherwise identical to `testServerInitiatedClosure` above, and it passes even without the `setClosePromise` reordering in `DefaultWebSocketConnection.initiateConnectionClosure()`.

The actual race the PR title describes (peer's close frame reaching the frame handler before the write listener publishes the close promise) is left untested, so a future revert of that reordering would go unnoticed. Worth adding a case where the client sends its own close frame concurrently with (or immediately before) the server initiating closure, asserting that the server's close future completes successfully and that `onClose` is *not* delivered to the connector listener.

TharmiganK (commented):
Claude PR Review Findings

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/websocket/DefaultWebSocketConnection.java
ThisaruGuruge: Good catch. Switched all three completion sites (write-failure listener here, channelInactive, and notifyCloseMessage) to tryFailure/trySuccess, and made closePromise volatile since it's now published off the event loop. e6971aa6e.

native/src/test/java/io/ballerina/stdlib/http/transport/websocket/server/WebSocketServerFunctionalityTestCase.java
ThisaruGuruge: Added testSimultaneousCloseFromBothEndpoints: client sends its own close frame (different code) right after the server independently initiates its own, without waiting to see the server's frame first. Forcing the exact write-failure-vs-echo interleaving deterministically would need test-only hooks into the channel, so this covers the independently-initiated-codes case rather than that exact byte-level race -- the tryFailure/trySuccess fix removes the risk regardless of which interleaving occurs.

native/src/test/java/io/ballerina/stdlib/http/transport/websocket/server/WebSocketServerFunctionalityTestCase.java
```diff
@@ -188,32 +188,41 @@ private WebSocketTestClient commandServerInitiatedClosure() throws URISyntaxExce
...
+        WebSocketTestClient client = new WebSocketTestClient();
+        client.handshake();
+
+        client.sendText("send-and-wait");
+        serverLatch.await(WEBSOCKET_TEST_IDLE_TIMEOUT, SECONDS);
+        client.sendCloseFrame(1000, null);
+        methodDoneLatch.await(WEBSOCKET_TEST_IDLE_TIMEOUT, SECONDS);
+
+        Assert.assertTrue(serverConnectorListener.getCloseFuture().isSuccess());
+        CloseWebSocketFrame closeFrame = client.getReceivedCloseFrame();
```
TharmiganK, line 229: **Race condition: the client-side close frame is read without ever waiting for it.**

Unlike `commandServerInitiatedClosure()` (which does `client.setCountDownLatch(clientLatch); ...; clientLatch.await(...)` before reading), this test never registers a client latch. The only synchronisation is `methodDoneLatch`, which is counted down on the **server's** event loop from the `closeFuture` listener. Nothing establishes a happens-before edge with the **client's** event loop, which is the thread that assigns `WebSocketTestClientFrameHandler.receivedCloseFrame` (a plain, non-`volatile` field).

So this is both a timing race (the client may not have decoded the server's close frame yet when the server-side round trip finishes) and a Java memory-model data race (the write may not be visible to the test thread even if it happened). The failure mode is an intermittent `assertNotNull(closeFrame)` failure on line 224 in CI.

Suggested fix — wait on the client too:

```java
CountDownLatch clientLatch = new CountDownLatch(1);
WebSocketTestClient client = new WebSocketTestClient();
client.handshake();
client.setCountDownLatch(clientLatch);

client.sendText("send-and-wait");
clientLatch.await(WEBSOCKET_TEST_IDLE_TIMEOUT, SECONDS);   // client got the server's close frame
serverLatch.await(WEBSOCKET_TEST_IDLE_TIMEOUT, SECONDS);
...
```

Also worth asserting the return value of the `await(...)` calls (`Assert.assertTrue(methodDoneLatch.await(...), "timed out")`) so a timeout reports as a timeout rather than as a confusing downstream NPE/assert.

native/src/test/java/io/ballerina/stdlib/http/transport/websocket/server/WebSocketServerFunctionalityTestCase.java
```diff
@@ -188,32 +188,41 @@ private WebSocketTestClient commandServerInitiatedClosure() throws URISyntaxExce
...
+    public void testSimultaneousCloseFromBothEndpoints() throws InterruptedException, URISyntaxException {
+        CountDownLatch serverLatch = new CountDownLatch(1);
+        serverConnectorListener.setReturnFutureLatch(serverLatch);
+        CountDownLatch methodDoneLatch = new CountDownLatch(1);
+        serverConnectorListener.setMethodDoneLatch(methodDoneLatch);
+        WebSocketTestClient client = new WebSocketTestClient();
+        client.handshake();
+
+        client.sendText("send-and-wait");
+        serverLatch.await(WEBSOCKET_TEST_IDLE_TIMEOUT, SECONDS);
```
TharmiganK: **This test does not actually cover the race that the fix addresses.**

The core production change in this PR is moving `frameHandler.setClosePromise(closePromise)` in `DefaultWebSocketConnection.initiateConnectionClosure` from inside the `writeAndFlush` listener to *before* the write, so that a close frame arriving before the write listener runs is recognised as part of the closing handshake instead of being reported to the app as a fresh remote-initiated close.

But `serverLatch` is counted down by `handleCloseFuture(...)` *synchronously, right after* `initiateConnectionClosure(...)` returns — i.e. after the close promise is already registered on the frame handler. By awaiting it on this line before calling `sendCloseFrame`, the test guarantees the ordering that the old code got wrong can never occur here. What remains exercised is only the removal of the status-code equality check (server 1001 vs. client 1000), which is exactly what the reworked `testSendAndReceiveDifferentStatusCode` above already covers — so the two tests are effectively redundant, and the promise-registration reordering ships untested.

To cover the real race you need the inbound close frame to reach the server handler before/concurrently with the local `writeAndFlush` — e.g. have the client send its close frame *before* triggering the server-side closure (so the server's `onMessage` handler for `send-and-wait` runs with a close frame already queued), or drive `initiateConnectionClosure` from a non-event-loop thread while the peer close frame is in flight.

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/websocket/WebSocketInboundFrameHandler.java
```diff
@@ -223,14 +223,7 @@ private void notifyCloseMessage(CloseWebSocketFrame closeWebSocketFrame) throws
...
         } else {
-            if (webSocketConnection.getCloseInitiatedStatusCode() != statusCode) {
-                String errMsg = String.format(
-                        "Expected status code %d but found %d in echoed close frame from remote endpoint",
-                        webSocketConnection.getCloseInitiatedStatusCode(), statusCode);
-                closePromise.setFailure(new IllegalStateException(errMsg));
-                return;
-            }
-            closePromise.setSuccess();
+            closePromise.trySuccess();
```
TharmiganK, line 226: Dropping the status-code check is the right call per RFC 6455 §5.5.1, but the **public interface contract now contradicts the implementation**.

`WebSocketConnection.initiateConnectionClosure(int, String)` and `initiateConnectionClosure()` (`native/src/main/java/io/ballerina/stdlib/http/transport/contract/websocket/WebSocketConnection.java`, ~lines 128-151) both still document:

> `ChannelFuture` will complete operation successfully **if and only if** it receives back a echoed close WebSocket frame from the remote endpoint **with the same status code as was written**.

That "same status code" guarantee is exactly the behaviour this line removes. Since `WebSocketConnection` is a published contract interface consumed outside this repo (notably `module-ballerina-websocket`), leaving the javadoc stale is how downstream implementers end up relying on a guarantee that no longer holds. Please update both overloads' javadoc to say the future completes successfully on receipt of *any* close frame from the remote endpoint, regardless of its status code.

TharmiganK (commented):
Claude PR Review Findings

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/websocket/DefaultWebSocketConnection.java
```diff
@@ -186,11 +185,9 @@ public ChannelFuture initiateConnectionClosure() {
...
         ctx.writeAndFlush(closeWebSocketFrame).addListener(future -> {
```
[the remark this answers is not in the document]
ThisaruGuruge: Same issue as #3781106304, already fixed in this branch — all three completion sites (write-failure listener, channelInactive, and notifyCloseMessage) now use trySuccess/tryFailure, and closePromise is volatile. See e6971aa6e.

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/websocket/WebSocketInboundFrameHandler.java
ThisaruGuruge: Good catch, thanks. Updated both javadoc overloads in WebSocketConnection.java to say the future completes on receipt of any close frame from the remote endpoint, regardless of its status code. e8b3aecad.

native/src/test/java/io/ballerina/stdlib/http/transport/websocket/server/WebSocketServerFunctionalityTestCase.java
ThisaruGuruge: Confirmed — receivedCloseFrame is a plain field and nothing established a happens-before edge with the client's event loop before this line read it. Added the missing clientLatch (set before sendText, awaited before reading the frame), matching the pattern in commandServerInitiatedClosure, and asserted the return value of every await(...) in this test so a timeout fails loudly instead of surfacing as a downstream NPE/assert. e8b3aecad.

native/src/test/java/io/ballerina/stdlib/http/transport/websocket/server/WebSocketServerFunctionalityTestCase.java
ThisaruGuruge: Fair point, and I agree the ordering here can't manifest the exact race: serverLatch counts down synchronously inside onMessage's still-executing call stack (right after setClosePromise, before any inbound frame from this test's client can physically round-trip back), so the local registration always wins.

I don't think this is practically fixable without new test-only hooks (e.g. an outbound handler that can delay/reorder the write completion relative to a queued inbound close frame) — this suite doesn't have anything like that today (no EmbeddedChannel-based tests), and I'd rather not add that kind of infrastructure on a 2.15.x patch branch for one race window.

Given that, I'd rather accept this as a coverage gap than a correctness gap: e6971aa6e made every completion site idempotent (trySuccess/tryFailure), so which side's event reaches the promise first no longer matters for correctness — only one of them will "win" and the other is a no-op instead of throwing. That was the same reasoning applied to the identical concern on the sibling thread. Leaving this open in case you see a lighter-weight way to force the interleaving that I'm missing.

TharmiganK (approved):

TharmiganK (approved):

PR #2668, opened 2026-08-17
Fix compiler validations skipped for service classes with multiple type qualifiers

### LINKED ISSUE #5085: HTTP compiler validations are not working for isolated service
**Description:**

When we define an isolated service class for interceptors, the compiler plugin validations are not performed. Because of this the program crashes at runtime.

**Steps to reproduce:**

Check the following code:
```ballerina
import ballerina/http;

isolated service class ResponseErrorInterceptor1 {
    *http:ResponseErrorInterceptor;
}

service class ResponseErrorInterceptor2 {
    *http:ResponseErrorInterceptor;
}
```

For the second service, there is a compiler error, but the first service does not show any errors

**Affected Versions:**

Ballerina Swan Lake Update 8

JaninduHK:
## Purpose

Fixes ballerina-platform/ballerina-library#5085

The HTTP compiler plugin skipped all interceptor validations for
`isolated service class`. Invalid interceptors compiled without error
and failed at runtime instead.

## Root cause

`HttpInterceptorServiceValidator` checked that *every* class type
qualifier was `service`:

```java
if (!tokens.stream().allMatch(token -> token.text().equals(Constants.SERVICE_KEYWORD))) {
    return;
}
```

For `isolated service class`, `classTypeQualifiers()` returns
`[isolated, service]`, so `allMatch` is false and the validator returns
before running any checks. Plain `service class` returns `[service]` and
validates normally.

The check should test for the *presence* of the `service` qualifier
rather than requiring it to be the only one, so `distinct`, `readonly`
and `client` are covered too.

The identical predicate also appeared in
`codemodifier/payload/HttpPayloadParamIdentifier#validateClassDefinition`,
which meant payload parameter identification was skipped for the same
classes. Both occurrences are fixed here.

## Before / after

Test source:

```ballerina
service class PlainInterceptor {
    *http:RequestInterceptor;
}

isolated service class IsolatedInterceptor {
    *http:RequestInterceptor;
}
```

Before only the non-isolated class is reported:

`ERROR [main.bal:(3:1,5:2)] RequestInterceptor must have a resource method`

After both are reported:

`ERROR [main.bal:(3:1,5:2)] RequestInterceptor must have a resource method`
`ERROR [main.bal:(7:1,9:2)] RequestInterceptor must have a resource method`

## Tests

Added `sample_package_52` covering `service`, `isolated service` and
`distinct service` class interceptors, plus
`CompilerPluginTest#testIsolatedInterceptorServiceClass`.

Verified the test fails without the fix (1 failure / 98) and passes with
it (98 / 98).

## Summary

- Updated HTTP compiler validation to recognize `service` in classes with multiple qualifiers.
- Added validation coverage for plain, isolated, and distinct interceptor service classes.
- Added payload annotation code-modifier coverage for regular and isolated service classes.
- Added an unreleased changelog entry.


compiler-plugin-tests/src/test/resources/ballerina_sources/sample_package_52/Ballerina.toml
```diff
@@ -0,0 +1,4 @@
+[package]
+org = "http_test"
+name = "sample_52"
+version = "0.1.0"
```
TharmiganK: Add a new empty line at the end of the file

compiler-plugin-tests/src/test/resources/ballerina_sources/sample_package_52/service.bal
```diff
@@ -0,0 +1,13 @@
...
+    *http:RequestInterceptor;
+}
+
+isolated service class IsolatedInterceptor {
+    *http:RequestInterceptor;
+}
+
+distinct service class DistinctInterceptor {
+    *http:RequestInterceptor;
+}
```
TharmiganK: Add a new empty line at the end of the file

TharmiganK:
@JaninduHK Can you please update the changelog as well: https://github.com/ballerina-platform/module-ballerina-http/blob/master/changelog.md

Add a note under fixed section

Also appreciate if you can send the similar fix for the branches - 2.15.x and 2.16.x

compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/codemodifier/payload/HttpPayloadParamIdentifier.java
```diff
@@ -153,7 +153,7 @@ private void validateClassDefinition(SyntaxNodeAnalysisContext syntaxNodeAnalysi
         if (tokens.isEmpty()) {
             return;
         }
-        if (!tokens.stream().allMatch(token -> token.text().equals(Constants.SERVICE_KEYWORD))) {
+        if (tokens.stream().noneMatch(token -> token.text().equals(Constants.SERVICE_KEYWORD))) {
```
TharmiganK, line 156: Can we add a test case for this as well?

compiler-plugin-tests/src/test/java/io/ballerina/stdlib/http/compiler/CompilerPluginTest.java
```diff
@@ -522,6 +522,17 @@ public void testInterceptorServiceObject() {
...
+    @Test
+    public void testIsolatedInterceptorServiceClass() {
+        Package currentPackage = loadPackage("sample_package_52");
+        PackageCompilation compilation = currentPackage.getCompilation();
+        DiagnosticResult diagnosticResult = compilation.diagnosticResult();
+        Assert.assertEquals(diagnosticResult.errorCount(), 3);
+        assertError(diagnosticResult, 0, "RequestInterceptor must have a resource method", HTTP_132);
+        assertError(diagnosticResult, 1, "RequestInterceptor must have a resource method", HTTP_132);
+        assertError(diagnosticResult, 2, "RequestInterceptor must have a resource method", HTTP_132);
+        }
```
TharmiganK: Minor formatting nit: this closing brace is indented 8 spaces instead of 4, inconsistent with every other test method in this file (e.g. `testReadonlyReturnTypes` right below it). Doesn't affect correctness, but likely to trip the module's checkstyle/formatting check.

compiler-plugin-tests/src/test/resources/ballerina_sources/sample_package_52/service.bal
TharmiganK: Per `ballerina-best-practices` Tier 3 (formatting/hygiene): file is missing a trailing newline at EOF, same as the new `Ballerina.toml` in this test resource package. `bal format` normally enforces a final newline — worth adding for consistency with the rest of the test resources.

TharmiganK (commented):
Claude PR Review Findings

JaninduHK:
Thanks for the review, addressed the trailing newlines on both `sample_package_52` files, the closing brace indentation, and added the changelog entry under Unreleased. I'd also left the Apache header off the `new service.bal`, so that's added now.

On the code modifier test: I tried a `sample_package_53` with an `isolated service class` implementing `*http:Service` and a resource with ambiguous payload parameters, the same shape that produces HTTP_151 in `sample_package_35`. It reports `errorCount() == 0` and `diagnosticCount() == 0` via both `getCompilation()` and `runCodeGenAndModifyPlugins()`, so there's nothing to assert against. Adding the package also broke `StaticCodeAnalyzerTest#testStaticCodeRulesWithAPI`, which asserts exact issue counts across the analyzed sources.

Rather than commit a test that passes vacuously, I've left it out of this push. Could you point me at the right fixture shape or harness for reaching `validateClassDefinition`? I'll add it as soon as I know where to hook in.

Also happy to send the 2.15.x and 2.16.x backports once this merges.

TharmiganK:
> On the code modifier test: I tried a `sample_package_53` with an `isolated service class` implementing `*http:Service` and a resource with ambiguous payload parameters, the same shape that produces HTTP_151 in `sample_package_35`. It reports `errorCount() == 0` and `diagnosticCount() == 0` via both `getCompilation()` and `runCodeGenAndModifyPlugins()`, so there's nothing to assert against. Adding the package also broke `StaticCodeAnalyzerTest#testStaticCodeRulesWithAPI`, which asserts exact issue counts across the analyzed sources.

@JaninduHK Let me check on this and get back at you

JaninduHK:
I found a possible way to test this behavior. We can add a compiler-plugin test using an `isolated service class` that implements `*http:Service`, run the code modifier, and assert that `@http:Payload` is added to the resource parameter.

I verified this approach locally: the test fails with the previous `!allMatch(...)` condition and passes with the corrected `noneMatch(...)` condition. I haven’t committed the test yet.

Would this be an acceptable approach? If so, I can add it to the PR.

TharmiganK:
> I found a possible way to test this behavior. We can add a compiler-plugin test using an `isolated service class` that implements `*http:Service`, run the code modifier, and assert that `@http:Payload` is added to the resource parameter.
> 
> I verified this approach locally: the test fails with the previous `!allMatch(...)` condition and passes with the corrected `noneMatch(...)` condition. I haven’t committed the test yet.
> 
> Would this be an acceptable approach? If so, I can add it to the PR.

Yes, that should be fine. 

But the diagnostics should also work, I think there might be some bug. I did not get time to check on this yet. If possible you can check that as well. You can create a separate issue for that.

```bal
import ballerina/http;

listener http:Listener listenerEp = new(8080);

type Person record {
    string name;
    int age;
};

service /api on listenerEp {
    // There is a diagnostic at p2 due to ambiquity
    resource function post test(Person p1, Person p2) {}
}

service class HttpService {
    *http:Service;
    // Missing diagnostic at p2 due to ambiquity
    resource function post test(Person p1, Person p2) {}
}

http:Service svc = service object {
    // Missing diagnostic at p2 due to ambiquity
    resource function post test(Person p1, Person p2) {}
};
```

IIUC, only the code modifier get engaged for service objects and classes and analyzers are not checking them

JaninduHK:
Thanks, I checked the diagnostic behavior and created a separate issue to track the missing ambiguity diagnostics for HTTP service classes and service objects:

https://github.com/ballerina-platform/ballerina-library/issues/9073

The code modifier test has been added to this PR and verifies that `@http:Payload` is inserted for an isolated service class.

TharmiganK (dismissed):
LGTM

TharmiganK (approved):
LGTM

TharmiganK:
@JaninduHK Can you send similar fixes to 2.15.x and 2.16.x branches as well

JaninduHK:
Sure, I’ll prepare the same fixes for both the `2.15.x` and `2.16.x` branches and send separate PRs. Thanks!

PR #2669, opened 2026-08-18
Fix large streamed messages being truncated when the application consumes them slowly

### LINKED ISSUE #9033: HTTP client intermittently aborts a large streamed response on a reused keep-alive connection — getByteStream() yields a prematurely closed stream ("Unexpected EOF in prolog")
### Area

Area/Connector

### Description

When consuming large HTTP responses (~124 MB XML) via response.getByteStream() and feeding the stream to xmldata:parseStream(), the client intermittently aborts the response read a few MB into the body and closes the TCP connection itself, after which the parser fails with:

`error: failed to parse xml: Unexpected EOF in prolog`

"In prolog" indicates the parser received effectively zero usable bytes — the stream was terminated before delivery, not truncated mid-document. Packet-level evidence (below) shows the client closes the connection first while the server is still writing; the server is not at fault.

**Scenario**

- HTTP/1.1 http:Client (default settings, keep-alive on) against a plain HTTP backend
- Sequential POSTs on the same pooled connection: ~19 MB XML request (sent chunked), ~124.5 MB XML response (Content-Length)
- 10 requests per processing cycle, cycles repeated back-to-back
response consumed with var body = check response.getByteStream(); then xmldata:parseStream(body, {enableConstraintValidation: false})

**Observed behavior**

- Failure rate ≈ 1 in 6 cycles with the default block size; same with getByteStream(512 * 1024)
- With a very large block size (getByteStream(200000 * 1024) ≈ 205 MB), failure rate jumps to ~5 in 6
- Both captured failures occurred after the same pooled connection had already served ~24–26 requests / ~3 GB cumulative download, and died ~2–3 MB into the next response

**Workaround**

Reading the full body with response.getBinaryPayload() and parsing via xmldata:parseBytes() is 100% stable under the identical workload (6/6 cycles, zero failures) — which isolates the defect to the streamed-consumption path, likely connection-pool/stream lifecycle interaction. (Disabling keep-alive is a plausible mitigation but was not tested.)

### Version

2201.13.4

### Environment Details (with versions)

- Ballerina 2201.13.4 (Swan Lake Update 13)
- JVM: Eclipse Temurin 21 (linux/arm64 container) and Homebrew OpenJDK on macOS — observed in containerized runs
- Backend: plain HTTP/1.1 server, Content-Length responses, no TLS
- Heap: reproduced at -Xmx2g with no memory pressure (zero full GCs in passing runs; failure is not an OOM)

TharmiganK:
## Purpose

> $subject

Fixes ballerina-platform/ballerina-library#9033

A large message that the application consumes more slowly than the configured `timeout` was torn down part way through, surfacing as a prematurely ended stream — for the reporter, an XML parse failure roughly one cycle in six on ~124 MB responses read through `getByteStream()`.

**Root cause.** Message bodies are moved on demand: inbound content is pulled off the socket only as the application drains what is already queued, and outbound content is written only as the peer accepts it. The idle timeout, however, measured nothing but read/write inactivity on the connection, so it could not distinguish *the peer has stopped responding* from *we have stopped asking*. In the second case the connection looks idle while the transfer is perfectly healthy, and acting on that truncates the message.

The same confusion existed in three places, with different thresholds, which is why the symptom looked erratic:

| Path | Back-pressure mechanism | Trips at |
|---|---|---|
| HTTP/1.1 inbound | `autoRead` disabled, reads re-armed as the queue drains | ~2 MB queued |
| HTTP/2 inbound | flow control window replenished only as the app consumes | 64 KB (default window) |
| HTTP/2 outbound | our frames queued in the remote flow controller | peer's window closes |

HTTP/1.1 outbound is not affected by the application's own pace — writes complete incrementally as the socket drains, so a slow producer keeps refreshing the timer. The mirror of the HTTP/2 outbound row is not covered, though: on an HTTP/1.1 connection a peer that stops reading is still treated as an unresponsive peer rather than as back-pressure. The spec now says so explicitly rather than implying the protection is symmetric across both directions.

**What changed.** Before failing a connection or stream, the timeout handlers now ask *why* it is quiet. If our own back-pressure explains the silence the countdown restarts instead; if the peer is genuinely unresponsive it fires on the same schedule as before. `DefaultListener` supplies a live suspension check rather than a cached flag, reports progress explicitly, and untracks itself when a message ends so a pooled connection is never left answering for a finished message. The HTTP/2 client and server handlers share the decision through `FlowControlStallTracker`.

Excusing back-pressure without limit would mean a connection could never be reclaimed from a hung application, or a stream from a peer that has stopped reading, so the allowance is bounded by the new `maxBackPressureStallTime` configurable and any progress restarts it.

The reporter has confirmed this resolves the issue against their workload.

## Examples

The allowance is a module level configurable, in seconds, defaulting to 300:

```toml
[ballerina.http]
maxBackPressureStallTime = 600
```

Two boundary values follow the convention `maxActiveConnections` already sets:

| Value | Effect |
|---|---|
| `-1` (negative) | back-pressure excused indefinitely, no cap |
| `0` | no reprieve; any inactivity is treated as an idle connection (previous behaviour) |
| `300` (default) | a transfer that makes no progress at all is treated as hung after 5 further minutes |

Any progress — a byte consumed, a frame read or written — restarts the span, so a merely slow transfer never approaches the cap. The span itself only begins once the connection's own `timeout` has already elapsed once for the transfer, since that is when the silence is first noticed: a transfer that never makes progress again is therefore reclaimed after `timeout + maxBackPressureStallTime`, not after the cap alone.

## Checklist

- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [x] Updated the spec

## Summary

- Prevent premature termination of large streamed HTTP messages during slow application consumption.
- Add back-pressure-aware timeout handling for HTTP/1.1 inbound, HTTP/2 inbound, and HTTP/2 outbound traffic.
- Add configurable `maxBackPressureStallTime`, with a default of 300 seconds. Negative values disable the limit, and zero disables the grace period.
- Track read suspension, flow-control stalls, application progress, and completed messages.
- Enforce stall limits for permanently stalled HTTP/2 streams.
- Add regression tests for slow large-message consumption and stall-limit enforcement.
- Update dependencies, changelog, and specification documentation.


native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/common/FlowControlStallTracker.java
```diff
@@ -0,0 +1,115 @@
...
+            return false;
+        }
+        if (isStreamBlockedByFlowControl(connectionSupplier.get(), streamId)) {
+            stallStartTimes.putIfAbsent(streamId, ticksInNanos());
+            return true;
+        }
+        if (stallStartTime != null) {
+            // The window has only just reopened; give the peer one full period to act before failing it.
+            stallStartTimes.remove(streamId);
+            return true;
```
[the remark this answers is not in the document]
TharmiganK: This is fixed now

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/common/BackPressureAwareIdleStateHandler.java
```diff
@@ -0,0 +1,194 @@
...
+                                  + "back-pressure", ctx.channel().id());
+                return;
+            }
+            // The stall start is deliberately left standing: until the application makes some progress, every
+            // further period should time out too rather than earn a fresh allowance.
+            LOG.debug("Idle timeout triggered on {}, inbound reads have been held up by application "
+                              + "back-pressure past the permitted stall time without any progress",
+                      ctx.channel().id());
+        }
+        super.channelIdle(ctx, evt);
```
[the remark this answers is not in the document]
TharmiganK: This is fixed now

ballerina/Dependencies.toml
```diff
@@ -5,7 +5,7 @@
 
 [ballerina]
 dependencies-toml-version = "2"
-distribution-version = "2201.13.3"
+distribution-version = "2201.13.5-20260805-130900-16f7d807"
```
niveathika, line 8: IS this due to bal build

TharmiganK: Yes, the local gradle build will update the lang version which is specified in the gradle.properties

TharmiganK:
@coderabbitai review

ballerina/Dependencies.toml
niveathika: The gradle.properties says 2201.13.5

TharmiganK: https://github.com/ballerina-platform/module-ballerina-http/blob/59aaa108443d1fd44edd892aa9947e79b6ea94df/gradle.properties#L4

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/common/BackPressureAwareIdleStateHandler.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    private static InboundReadState getOrCreateState(Channel channel) {
+        Attribute<InboundReadState> attribute = channel.attr(INBOUND_READ_STATE);
+        InboundReadState state = attribute.get();
+        if (state == null) {
+            InboundReadState newState = new InboundReadState();
+            InboundReadState existingState = attribute.setIfAbsent(newState);
+            state = existingState != null ? existingState : newState;
+        }
```
ThisaruGuruge: I'd invert the condition here:


```suggestion
        if (state != null) {
            return state;
        }
        InboundReadState newState = new InboundReadState();
        InboundReadState existingState = attribute.setIfAbsent(newState);
        return existingState != null ? existingState : newState;
```

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/common/BackPressureAwareIdleStateHandler.java
```diff
@@ -0,0 +1,191 @@
...
+        long delay = Math.max(remaining, MIN_RECHECK_NANOS);
+        state.setStallLimitRecheckTask(ctx.channel().eventLoop().schedule(() -> {
+            if (!ctx.isRemoved()) {
+                try {
+                    channelIdle(ctx, evt);
+                } catch (Exception e) {
+                    LOG.debug("Error while re-checking the back-pressure stall limit on {}", ctx.channel().id(), e);
+                }
+            }
```
ThisaruGuruge: Ditto

```suggestion
            if (ctx.isRemoved()) {
                return
            }
            try {
                channelIdle(ctx, evt);
            } catch (Exception e) {
                LOG.debug("Error while re-checking the back-pressure stall limit on {}", ctx.channel().id(), e);
            }
```

TharmiganK: Fixed via [f83cf37](https://github.com/ballerina-platform/module-ballerina-http/pull/2669/commits/f83cf373492fa3e1ba88084543c0b23a64083df8)

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/common/BackPressureAwareIdleStateHandler.java
TharmiganK: [f83cf37](https://github.com/ballerina-platform/module-ballerina-http/pull/2669/commits/f83cf373492fa3e1ba88084543c0b23a64083df8)

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/sender/http2/Http2ClientTimeoutHandler.java
```diff
@@ -82,7 +85,8 @@ private void setTimerTask(ChannelHandlerContext ctx, int streamId, OutboundMsgHo
     }
 
     public void createTimerTask(ChannelHandlerContext ctx, int streamId, long timeOut, boolean expectContinue) {
-        this.idleTimeNanos = timeOut;
+        // timeOut is in milliseconds; idleTimeNanos is read as nanoseconds everywhere else in this class.
+        this.idleTimeNanos = TimeUnit.MILLISECONDS.toNanos(timeOut);
```
ThisaruGuruge: `idleTimeNanos` is an instance field shared by every stream multiplexed on this connection, but `createTimerTask` is invoked per-stream — including for `Expect: 100-continue`, with `socketIdleTimeout/5`. That call overwrites the shared budget for every other in-flight stream on the same connection. Example: stream A is mid-transfer and healthy against a 60s `socketIdleTimeout`; stream B sends `Expect: 100-continue`, which sets `idleTimeNanos` to 12s. If A's idle/recheck logic reads `idleTimeNanos` while B is still pending, A gets reset/timed out against the corrupted 12s budget despite being well within its real 60s allowance.

Consider keying this per-stream (e.g. in the stream's own `IdleTimeoutTask`/map entry) instead of on the shared handler field.

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/common/BackPressureAwareIdleStateHandler.java
```diff
@@ -0,0 +1,193 @@
...
+            try {
+                channelIdle(ctx, evt);
+            } catch (Exception e) {
+                LOG.debug("Error while re-checking the back-pressure stall limit on {}", ctx.channel().id(), e);
+            }
+        }, delay, TimeUnit.NANOSECONDS));
+    }
+
+    private boolean isCausedByBackPressure(InboundReadState state) {
+        return state.isReadSuspended() || ticksInNanos() - state.getLastResumeTimeNanos() < idleTimeoutNanos;
```
ThisaruGuruge: This handler runs earlier in the pipeline than Netty's own idle-state tracking, so `lastResumeTimeNanos` is set a few microseconds ahead of Netty's internal last-read baseline for the same read. On a lightly loaded event loop, Netty's own idle check can fire with only a tiny overshoot past `idleTimeoutNanos`, which can make `ticksInNanos() - lastResumeTimeNanos < idleTimeoutNanos` true even with zero real back-pressure. That grants a genuinely idle (not back-pressured) connection an undeserved `maxBackPressureStallTime` grace period (300s default) before it's reclaimed.

docs/spec/spec.md
```diff
@@ -2076,6 +2077,21 @@ public class Response {
 
 The header and the payload manipulation can be done using the functions associated to the response.
 
+### 6.1. Message body streaming and back-pressure
+
+Message bodies are moved on demand rather than buffered in full. An inbound body is read from the connection as the application consumes what has already been delivered, and an outbound body is written as the remote endpoint accepts it. An application that consumes or produces a body slowly therefore holds the transfer still, and no data moves on the connection while that lasts.
```
ThisaruGuruge, line 2082: This reads as claiming symmetric protection for both directions ("consumes or produces a body slowly"), but from the diff it looks like only the inbound/read path is wired up here for HTTP/1.1 (`BackPressureAwareIdleStateHandler` + `DefaultListener`). HTTP/2 has `FlowControlStallTracker` for the outbound side, but I don't see an HTTP/1.1 outbound equivalent in this PR. Worth confirming a slow-writing HTTP/1.1 peer is actually covered, or narrowing this wording to the inbound case.

native/src/main/java/io/ballerina/stdlib/http/transport/message/DefaultListener.java
```diff
@@ -43,32 +45,60 @@ public DefaultListener(ChannelHandlerContext ctx) {
...
     @Override
     public void onRemove(HttpContent httpContent) {
+        ChannelHandlerContext currentCtx = this.ctx;
+        if (currentCtx == null) {
+            this.cumulativeByteQuantity.addAndGet(-(httpContent.content().readableBytes()));
+            return;
+        }
+        // Recorded before the reduced count is published, so that whoever observes the lower count also
+        // observes the refreshed resume time instead of a stale one that could time the peer out early.
+        BackPressureAwareIdleStateHandler.inboundReadsResumed(currentCtx.channel());
```
ThisaruGuruge: `currentCtx` is snapshotted from `this.ctx` before calling `inboundReadsResumed`. On a pooled/reused channel, if this call is delayed past the point where the message completes (`ctx` nulled) and a *new* message has already started tracking on the same channel, this stale call lands on the new message's shared `InboundReadState` and resets its stall clock (`resumed()` cancels the pending recheck task and clears `stallStartTimeNanos`). Since it only resets rather than extends, the effect is extra leniency (a stalled message might survive a bit past `maxBackPressureStallTime`) rather than a wrongful termination — but worth confirming this interleaving can't happen given the threading model here.

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/common/BackPressureAwareIdleStateHandler.java
```diff
@@ -0,0 +1,193 @@
...
+        // previous message, so every message this handler is installed for starts with a clean countdown.
+        getOrCreateState(ctx.channel()).reset();
+        super.handlerAdded(ctx);
+    }
+
+    @Override
+    protected void channelIdle(ChannelHandlerContext ctx, IdleStateEvent evt) throws Exception {
+        InboundReadState state = ctx.channel().attr(INBOUND_READ_STATE).get();
+        if (state != null && isCausedByBackPressure(state)) {
+            long stallStart = state.recordStall();
```
ThisaruGuruge, line 118: `recordStall()` is only reached once `channelIdle` fires for the first time, which itself requires `idleTimeoutNanos` of prior silence (Netty's own idle check). So the effective worst-case time to reclaim a stalled connection is `idleTimeoutNanos + maxBackPressureStallTime`, not `maxBackPressureStallTime` alone as the spec (line 2086) and PR description suggest. Worth clarifying the docs, or starting the stall clock earlier if a tighter bound is actually intended.

native/src/test/java/io/ballerina/stdlib/http/transport/BackPressureStallLimitTestCase.java
```diff
@@ -0,0 +1,157 @@
...
+        assertTrue(totalRead < RESPONSE_SIZE,
+                   "Response body was not truncated even though the consumer stalled past the configured "
+                           + "maxBackPressureStallTime cap");
+        assertTrue(reclaimDurationMillis < RECLAIM_DEADLINE_MILLIS,
+                   "Connection was not reclaimed until " + reclaimDurationMillis + "ms into the stall, which is "
+                           + "close to twice the socket idle timeout rather than close to the configured "
+                           + "maxBackPressureStallTime cap");
+    }
+
+    @AfterClass(timeOut = CLEAN_UP_TIME_OUT)
```
ThisaruGuruge: `@BeforeClass` sets a global static (`Util.setMaxBackPressureStallTime`) and starts a server that can fail to bind (e.g. a port still in TIME_WAIT from a prior class). Since this `@AfterClass` doesn't set `alwaysRun = true`, TestNG skips it if `@BeforeClass` throws, leaving the shortened stall cap applied JVM-wide for the rest of the suite — which can surface as unrelated flakiness in later test classes. `Http2ClientBackPressureStallLimitTestCase` has the same gap.

ThisaruGuruge (commented):
Reviewed for correctness/functionality bugs only (no nitpicks). Left inline comments on 6 issues below. One more doesn't fit as an inline comment because the file isn't touched by this PR:

**`native/src/main/java/io/ballerina/stdlib/http/transport/message/PassthroughBackPressureListener.java`** (unchanged by this PR) — HTTP/1.1 passthrough back-pressure (`onUnWritable`/`onWritable` toggling `autoRead`) never registers with the new `BackPressureAwareIdleStateHandler` tracker; `DefaultListener.isReadSuspended()` is the only signal that handler consults. For a passthrough/gateway service where the downstream write side goes unwritable, upstream reads get suspended without the tracker knowing about it, and once `socketIdleTimeout` elapses with no reads the upstream connection is torn down mid-transfer — the same class of bug this PR fixes for the non-passthrough case.

The two comments on `Http2ClientTimeoutHandler.java` and `PassthroughBackPressureListener.java`/passthrough path look like they let the exact bug this PR targets (premature termination of a healthy transfer) still occur via paths this PR doesn't cover.

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/sender/http2/Http2ClientTimeoutHandler.java
TharmiganK: Good catch, and the A/B scenario is exactly right. Fixed in cf081b239: `idleTimeNanos` is no longer an instance field. Each `IdleTimeoutTask` captures the budget it was created with, and the handler keeps only `defaultIdleTimeNanos` for the streams timed from `onStreamInit`, so `getNextDelay()` and the flow-control recheck both read the stream's own copy. A 100-continue wait on one stream can no longer shorten another's.

`Http2ClientTimeoutHandlerConcurrentStreamTest` (3600809bd) reproduces your scenario directly — a healthy stream fed regular progress against its own budget while a second stream waits on `socketIdleTimeout / 5`; it fails on the pre-fix code with a `resetStream` on the healthy stream.

One related thing while I was here: 5364617a0 floors the per-stream value with `MIN_TIMEOUT_NANOS`, the way the default already was, so `socketIdleTimeout / 5` cannot round down to a budget that has expired before it is set.

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/common/BackPressureAwareIdleStateHandler.java
TharmiganK: Fixed, in three steps.

b2d515d73 replaced the resume timestamp in that check with a raw read time recorded in `channelReadComplete` — the same method the superclass updates its own baseline in, and recorded just before it, so it can never run ahead of it.

That left a narrower version of the same problem, which I found while re-reviewing: `channelReadComplete` also fires for reads that carry no decoded message — a partial chunk, a partial TLS record, a read that returned nothing — and the superclass deliberately does not count those (it gates on the `reading` flag set in `channelRead`). So an untracked connection could still be excused. Measured on an `EmbeddedChannel` with a 150 ms timeout: one bare `channelReadComplete` at 100 ms pushed the first idle event from ~150 ms to ~305 ms, and from there it picked up the stall allowance. 5364617a0 gates the check on a message actually being tracked, and f342164d5 restricts the resume timestamp (see below) to resumes that end a suspension the idle check has run into. `testUntrackedChannelIsNotExcusedByAReadThatCarriedNoMessage` and `testProgressWithoutASuspensionDoesNotExtendTheCountdown` both fail on the pre-fix code.

The reason the resume timestamp is back at all is the mirror bug that dropping it introduced: with only the raw read time — stale by definition while reads are suspended — a peer was timed out moments after the application drained and reads were reissued, getting whatever was left of the period instead of a fresh one. Measured 176 ms of a 1000 ms budget in the same harness. `testPeerGetsAFullPeriodAfterReadsResume` pins that, and it is now the only case that refreshes the timestamp.

docs/spec/spec.md
TharmiganK: You're right — only the inbound direction is wired up for HTTP/1.1. Narrowed the wording in b2d515d73: the paragraph now states that for HTTP/2 the exclusion covers both directions of a stream, while for HTTP/1.1 only the inbound direction participates, and that a peer which reads a response or request body slowly from an HTTP/1.1 connection is not yet distinguished from one that has genuinely stopped responding.

The PR description had the same problem from the other side — it claimed HTTP/1.1 outbound was simply "not affected", which reads as "no gap". It now separates the two: the application's own write pace is genuinely not affected (writes complete incrementally as the socket drains), but the mirror of the HTTP/2 outbound row is not covered.

native/src/main/java/io/ballerina/stdlib/http/transport/message/DefaultListener.java
TharmiganK: It can happen — `onRemove` runs on the application thread, so nothing orders it against the event loop starting a new message on a pooled channel. Rather than rely on reasoning about the interleaving, the call is now guarded explicitly: `trackInboundReads` takes the owner, and `inboundReadsResumed` is ignored when the caller is no longer the tracked owner (b2d515d73, with a regression test in e4858e785).

5364617a0 extends the same guard to `untrackInboundReads`, which is the more destructive of the two: a late `resumeReadInterest()` from a superseded listener would otherwise clear `readSuspended` for the message tracked in its place, leaving that message's back-pressure invisible to the idle timeout — a wrongful termination rather than extra leniency. The same commit gives a newly tracked message a clean stall clock, so a message that ends without ever untracking (an aborted one, on the listener side where the handler is installed once per connection) cannot leave its clock behind for the next message to be reclaimed against. Both are covered by regression tests that fail on the pre-fix code.

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/common/BackPressureAwareIdleStateHandler.java
TharmiganK: Correct, the effective bound is `idleTimeoutNanos + maxBackPressureStallTime`.

That looser bound is the intended behaviour: the stall clock can only start once the silence has actually been noticed, and starting it earlier would mean maintaining a stall clock on every connection rather than on the ones the idle check has run into. So I've clarified the docs rather than tightening the bound. The spec paragraph now says a transfer is treated as stalled once the connection's own `timeout` has already elapsed once for it, is excused for up to `maxBackPressureStallTime` further seconds, and is therefore reclaimed after `timeout + maxBackPressureStallTime` rather than after the cap alone (b2d515d73).

The PR description carried the same wrong bound in its table of boundary values, and is now updated to match.

native/src/test/java/io/ballerina/stdlib/http/transport/BackPressureStallLimitTestCase.java
TharmiganK: Fixed in b2d515d73. Both `BackPressureStallLimitTestCase` and `Http2ClientBackPressureStallLimitTestCase` now use `@AfterClass(alwaysRun = true)`, so the shortened cap is restored even when `@BeforeClass` throws. The shutdown calls in the cleanup are null-guarded as well, so the restore is not itself lost to an NPE when setup failed partway through.

TharmiganK:
@ThisaruGuruge I've replied to and resolved the six inline threads. The seventh finding from your review — the one on `PassthroughBackPressureListener`, which had no inline anchor because the file was untouched — is fixed in cf081b239:

`PassthroughBackPressureListener` now takes the `DefaultListener` reading the incoming channel alongside the context, and tells it when the downstream leg goes unwritable and writable again. `DefaultListener.isReadSuspended()` — the one signal the idle handler consults — reports that suspension too, so a passthrough or gateway service whose downstream write side blocks is excused the same way the internal queue cap already was, and the resume restarts the countdown. `PassthroughBackPressureListenerTest` (3600809bd) covers both the excused case and a control case where the suspension is applied without notifying the listener.

Your assessment that the two findings let the original bug through on paths this PR did not cover was correct in both cases — thanks for catching them.

The spec and the PR description have also been corrected on the two documentation points (the direction asymmetry, and the `timeout + maxBackPressureStallTime` bound).

ThisaruGuruge (approved):

PR #2686, opened 2026-09-03
Update ballerinaLangVersion

chiranSachintha:
Updated ballerinaLangVersion to 2201.13.6

## Checklist
- [ ] Linked to an issue
- [ ] Updated the changelog
- [ ] Added tests
- [ ] Updated the spec

Updated `ballerinaLangVersion` in `gradle.properties` from a snapshot version to the `2201.13.6` release.


gradle.properties
```diff
@@ -1,7 +1,7 @@
 org.gradle.caching=true
 group=io.ballerina.stdlib
 version=2.17.0-SNAPSHOT
-ballerinaLangVersion=2201.13.5-20260805-130900-16f7d807
+ballerinaLangVersion=2201.13.6
```
TharmiganK, line 4: Why do we need this change? @daneshk @ThisaruGuruge I thought we were reverting lang version change?

daneshk: we need to lang api introduced in the newer version. I had a chat with Chiran and Asma, from language side, they are handling the previous version incompatibility.

chiranSachintha: Because we added a new API in 13.6 lang, we need to update this to use it in http. If the user tries to use the new stdlib version with a previous lang release, they will get a "not supported" error.

TharmiganK: This is fully tested right? It should not break with the older distributions.

TharmiganK: Should it be an error i.e. are we not generating the jar? Also can you point me to the location where we return an unsupported error?

chiranSachintha: https://github.com/ballerina-platform/module-ballerina-http/pull/2689

We are given warning as
```
WARNING [:(1:1,1:1)] The Ballerina version is not supported for endpoints.yaml. Try using Ballerina 2201.13.6 or above
```

TharmiganK (approved):

PR #2707, opened 2026-09-11
Migrate to Netty 4.2.18.Final

### LINKED ISSUE #9152: Migrate the standard library to Netty 4.2 before the 4.1 EOL
### Why now

Netty has [announced that 4.1 reaches End-of-Life on July 1st, 2027](https://netty.io/news/2026/09/09/4-1-EOL-announcement.html):

> Today we are announcing that Netty 4.1 will be End-of-Life on July 1st, 2027. After this date, we will no longer support Netty 4.1, and we will make no further releases in this series.

"No further releases" includes security releases. Netty CVEs have driven a steady stream of patch bumps across our modules, so losing the 4.1 security stream, rather than any new 4.2 feature, is what makes this urgent. Feature work has been 4.2-only since April 2025; 4.1 has received bug and security fixes only.

Current versions: **4.2.18.Final** (latest 4.2.x) and **4.1.138.Final** (latest 4.1.x), both released 09-Sep-2026. netty-tcnative pairs at **2.0.84.Final** for both.

### Why this has to be coordinated

From the [Netty 4.2 migration guide](https://netty.io/wiki/netty-4.2-migration-guide.html):

> Netty 4.1 and 4.2 cannot co-exist on the class path, so you need to ensure you do the upgrade in a single step.

Our modules pin and ship Netty jars independently, and several end up on the same application classpath. `ballerina/websocket` declares a **runtime** dependency on `ballerina/http`, so any application importing `websocket` gets both dependency sets. The same applies to any application combining `http` with `grpc`, `tcp`, `udp`, `redis`, `aws.s3` and so on.

Comparing the shipped artifact lists of `http` (migrated) against `aws.s3` (on 4.1) shows two distinct failure modes:

**1. Version conflict — visible to resolution.** Eight artifacts are shipped by both, at 4.1.137 and 4.2.18: `netty-buffer`, `netty-codec-http`, `netty-codec-http2`, `netty-common`, `netty-handler`, `netty-resolver`, `netty-transport`, `netty-transport-native-unix-common`. Resolution picks one version, so one of the two libraries runs against a Netty major it was not built for.

**2. Class collision — invisible to resolution.** `aws.s3` ships `netty-codec` 4.1.137; a migrated module ships `netty-codec-base` 4.2.18. Different artifactIds, so no version conflict is detected, yet the two jars carry **95 identically named classes** (`DefaultHeaders`, `MessageAggregator`, `ByteToMessageDecoder`, `Base64`, …), 15 of which have different member sets. Class loading order decides which wins.

The second is the more dangerous one. The public API of the colliding classes is largely unchanged, so the application will most likely link and appear to work, failing later and non-deterministically depending on jar order. That is harder to diagnose than a clean startup failure, and it is why this cannot be done module by module across releases.

### Modules that need migration

Modules that pin Netty directly and own their usage — these can be migrated on our own schedule:

| Module | Label | Netty | tcnative | Ships `netty-codec` |
|---|---|---|---|---|
| `module-ballerina-http` | `module/http` | 4.1.137.Final | 2.0.83.Final | yes |
| `module-ballerina-websocket` | `module/websocket` | 4.1.137.Final | — | yes |
| `module-ballerina-grpc` | `module/grpc` | 4.1.137.Final | 2.0.83.Final | yes |
| `module-ballerina-tcp` | `module/tcp` | 4.1.137.Final | — | yes |
| `module-ballerina-udp` | `module/udp` | 4.1.137.Final | — | no |
| `module-ballerinax-redis` | `module/redis` | 4.1.137.Final | — | yes |
| `module-ballerinax-jaeger` | `module/jaeger` | 4.1.135.Final | — | no |

Modules where Netty comes in through a third-party SDK. None of these have `io.netty` usage in their own Java source — Netty is shipped only to satisfy the SDK, so the SDK dictates the version:

| Module | Label | Netty | SDK | Status |
|---|---|---|---|---|
| `module-ballerinax-azure-cosmosdb` | `module/cosmosdb` | 4.1.100.Final | `reactor-netty` 1.1.13 | **Can be migrated** — see below |
| `module-ballerinax-aws.s3` | `module/aws-s3` | 4.1.137.Final | AWS SDK v2 `netty-nio-client` 2.42.0 | **Needs validation** |
| `module-ballerinax-asb` | `module/azure-servicebus` | 4.1.137.Final | `azure-core-http-netty` 1.13.6 | **Needs validation** |

**`cosmosdb` can be done.** Reactor Netty supports Netty 4.2 from its 1.3.x line: `reactor-netty-http` 1.3.0 depends on Netty 4.2.7.Final, and 1.4.0-M1 on 4.2.17.Final. The module is pinned to 1.1.13 (Netty 4.1.101), so this is an SDK version upgrade we control rather than an upstream wait.

**`aws.s3` needs validation.** The AWS SDK has not moved: even the latest release, 2.54.16, pins `netty.version = 4.1.137.Final`, so running it on 4.2 is untested by AWS. There is also a known runtime hazard — `software.amazon.awssdk.http.nio.netty.internal.utils.ChannelResolver` performs `instanceof NioEventLoopGroup` / `instanceof EpollEventLoopGroup` and otherwise throws `IllegalArgumentException("Unknown event loop group : ")`. In practice this may not trigger, because both classes still exist in Netty 4.2 (deprecated, now extending `MultiThreadIoEventLoopGroup`) and `SdkEventLoopGroup` constructs a `NioEventLoopGroup` itself, so the SDK's own default satisfies its own check. It would only fail if a 4.2-native group were handed to the SDK. This needs to be confirmed by running the module's tests against 4.2 rather than assumed either way.

**`asb` needs validation.** `azure-core-http-netty` delegates to Reactor Netty, so it should inherit 4.2 support in principle, but it has not shipped a release that does: the latest, 1.16.7, still uses reactor-netty 1.2.18 and pins Netty 4.1.137.Final. There are also community reports of runtime problems with Netty 4.2's default adaptive allocator when used alongside the Azure client, with the mitigation being to pin `-Dio.netty.allocator.type=pooled`. We have not reproduced that, so it should be treated as an open risk to verify, not a settled fact.

### What changes in 4.2

**Packaging.** Because we ship flat, pinned jar lists rather than resolving transitively, a bare version bump produces a broken package rather than a failing build:

- `netty-codec` is now an **empty aggregator** — `netty-codec-4.2.18.Final.jar` contains zero classes. Every module shipping it must replace it with `netty-codec-base`, plus `netty-codec-compression` where `ZlibCodecFactory`, `StandardCompressionOptions` or similar are used. `netty-codec-base` keeps the `io.netty.codec` JPMS module name, so `module-info.java` needs no change for it, but `netty-codec-compression` needs a new `requires` clause.
- `OcspClientHandler` moved out of `netty-handler` into a new `netty-handler-ssl-ocsp` artifact, whose module hard-requires `io.netty.resolver.dns` (which requires `io.netty.codec.dns`). Among our modules only `http` uses it.
- Netty 4.2 jars are real JPMS modules and no longer carry `Automatic-Module-Name`.
- `jboss-marshalling` and `protobuf-java` may become droppable where they were only present for the old monolithic `netty-codec`.
- `aws.s3` ships `netty-nio-client`, which requires `netty-resolver-dns`, but that artifact is missing from its `Ballerina.toml`. The migration should fix this.

**Behaviour — needs testing, not just compiling:**

- **Client TLS hostname verification is on by default.** `SslContextBuilder`'s `endpointIdentificationAlgorithm` now defaults to `HTTPS` where 4.1 defaulted to `null`. Any module that disables verification by *omission* will silently start verifying. The fix is to set it explicitly through `SslContextBuilder.endpointIdentificationAlgorithm(...)`, available since 4.1.112 and therefore landable ahead of the bump.
- **The adaptive allocator is the new default**, replacing pooled. There are open upstream reports of direct-memory regressions ([#15850](https://github.com/netty/netty/issues/15850), [#17107](https://github.com/netty/netty/issues/17107), [#15767](https://github.com/netty/netty/issues/15767)). The guide recommends upgrading with `-Dio.netty.allocator.type=pooled` and flipping separately once load-tested.
- **HTTP/2 header value validation is on by default** as of 4.2.18, and chunked-must-be-last is validated regardless of HTTP version.
- The pipeline call-stack was flattened, which can change behaviour where the pipeline is modified concurrently or uses child executors.
- `NioEventLoopGroup` and the other transport-specific groups are deprecated in favour of `MultiThreadIoEventLoopGroup` with an `IoHandlerFactory`. They still exist and are source-compatible, so this is a follow-up rather than a blocker.

**Native image.** Each split codec artifact ships its own `META-INF/native-image` metadata, so changing the jar set changes what reaches the image. 4.2 also adds classes that must be registered — `io.netty.handler.ssl.JdkDelegatingPrivateKeyMethod` needs `--initialize-at-run-time`, and a JVM build will not reveal it. Every module with native-image config needs a real GraalVM build, not just a green test suite.

**Security.** Target **4.2.17 or later**. [GHSA-p85m-gvr3-788c](https://github.com/netty/netty/security/advisories/GHSA-p85m-gvr3-788c) (high) disables hostname verification on the OpenSSL client path for `netty-handler` 4.2.0–4.2.16, affecting any module using `SslProvider.OPENSSL`.

### Suggested sequence

1. Validate `aws.s3` and `asb` against Netty 4.2, since whether they can move determines if a single-step migration is achievable at all.
2. Upgrade `cosmosdb` to Reactor Netty 1.3.x, which brings it onto Netty 4.2.
3. Bring every module to the latest 4.1.x (4.1.138.Final) and tcnative 2.0.84.Final, so any fallout is isolated from the 4.2 move.
4. Make endpoint identification explicit wherever client TLS contexts are built — doable on 4.1 and independently verifiable.
5. Move the modules together to 4.2.18.Final, reworking the artifact lists, and release them into the same distribution.
6. Verify each with a real GraalVM native-image build, not only JVM tests.
7. Decide the allocator separately, on load-test evidence, keeping `pooled` pinned until then.

Sub-issues to be raised per module.

TharmiganK:
## Purpose

Migrate the HTTP module from Netty 4.1 to Netty 4.2.18.Final.

[Netty 4.1 reaches End-of-Life on July 1st, 2027](https://netty.io/news/2026/09/09/4-1-EOL-announcement.html), after which there are no further releases in the series, security fixes included. This module has taken a steady stream of Netty CVE bumps, so losing that stream is the driver rather than any new 4.2 feature.

Fixes: ballerina-platform/ballerina-library#9152

> [!IMPORTANT]
> **This should not be merged on its own.** Netty states that [4.1 and 4.2 cannot co-exist on the class path](https://netty.io/wiki/netty-4.2-migration-guide.html). `ballerina/websocket` has a runtime dependency on `ballerina/http` and ships its own Netty 4.1 jars, as do `grpc`, `tcp`, `udp` and others, so an application combining them would end up with eight artifacts at two incompatible versions plus 95 identically named classes from `netty-codec` 4.1 and `netty-codec-base` 4.2 — a collision that dependency resolution cannot detect, because the artifact IDs differ. See ballerina-platform/ballerina-library#9152 for the coordination plan.

## Approach

Structured as separate commits, following the [Netty 4.2 migration guide's](https://netty.io/wiki/netty-4.2-migration-guide.html) recommended process, so each step is reviewable and bisectable on its own.

**1. Move to the latest 4.1.x first** — Netty 4.1.138.Final and netty-tcnative 2.0.84.Final. The guide's first step, so any fallout from the patch bump is isolated from the 4.2 move. `netty-bom` pins tcnative 2.0.84.Final for both 4.1.138 and 4.2.18, which lets the tcnative bump and its native-image revalidation land here rather than muddying the 4.2 commit.

**2. Configure endpoint identification explicitly on client SSL contexts.** Host name verification was applied additively: the endpoint identification algorithm was set on the `SSLEngine` only when `verifyHostName` was true, leaving the Netty `SslContext` default untouched otherwise. On 4.1 that default is `null`, so disabling worked by omission. Netty 4.2 changes `SslContextBuilder`'s default to `HTTPS`, which would silently keep verifying for users who set `verifyHostName = false` on every client path that goes through a Netty `SslContext` rather than a raw JDK `SSLContext`. All three client builders now set it explicitly, so our configuration is authoritative in both directions. `SslContextBuilder.endpointIdentificationAlgorithm` exists since 4.1.112, so this lands ahead of the bump and is verifiable on both versions.

**3. Migrate to Netty 4.2.18.Final.** Reworks the shipped artifact set:

- `netty-codec` is now an empty aggregator — `netty-codec-4.2.18.Final.jar` contains zero classes. Since this package pins a flat, non-transitive jar list, a bare version bump would ship that empty jar and fail at runtime. Replaced with `netty-codec-base` (which keeps the `io.netty.codec` JPMS module name, so `module-info.java` needs no change for it) and `netty-codec-compression`, which `ZlibCodecFactory` and `StandardCompressionOptions` need and which `netty-codec-http`/`http2` only declare as `requires static`.
- `OcspClientHandler` moved out of `netty-handler` into `netty-handler-ssl-ocsp`. `OCSPStaplingHandler` extends it, and that module hard-requires `io.netty.resolver.dns`, which requires `io.netty.codec.dns`, so all three artifacts are added. The latter two exist only to satisfy JPMS resolution; no code calls them.
- Dropped `jboss-marshalling` and `protobuf-java`. They were only present because the 4.1 monolithic `netty-codec` referenced the marshalling and protobuf codecs; those now live in separate artifacts we do not ship, and no class under `io.netty.handler.codec.{marshalling,protobuf}` remains in the packaged set. `lz4-java` stays, since `netty-codec-compression` still carries the Lz4 codecs.

**4. Register `JdkDelegatingPrivateKeyMethod` for native image.** GraalVM image generation fails with `Class initialization of io.netty.handler.ssl.JdkDelegatingPrivateKeyMethod failed`. The class is new in 4.2 and, like the `OpenSslPrivateKeyMethod` entries beside it, must be initialized at run time.

**5. Bump the package version to 2.18.0-SNAPSHOT**, since the shipped platform dependency set changes.

## Verification

Netty API compatibility was checked by extracting all 208 distinct `io.netty` imports in the repo and resolving each against the actual class listing of the 4.2.18 jars. Exactly one did not resolve — `OcspClientHandler`, addressed above. The low-level HTTP/2 API this module subclasses heavily (`Http2ConnectionHandler`, `AbstractHttp2ConnectionHandlerBuilder`, `Http2EventAdapter`, `Http2FrameLogger`, `CompressorHttp2ConnectionEncoder`) is signature-identical in 4.2, so no transport code changed.

| Check | Result |
|---|---|
| `:http-native:compileJava`, `compileTestJava`, `:http-test-utils:compileJava` | pass |
| `checkstyleMain`, `spotbugsMain` | pass |
| `http-security-tests` | 120 passing, 0 failing |
| `http2-tests` | 162 passing, 0 failing |
| All 208 `io.netty` imports resolve against the shipped jar set | pass |
| Every `io.netty` class in `native-image.properties` / `jni-config.json` / `reflect-config.json` present in 4.2 | pass |

**Not yet verified, and needed before merge:**

- The full `:http-native:test` suite.
- A green GraalVM native-image build. The `JdkDelegatingPrivateKeyMethod` fix above was found by a native-image run but has not yet been confirmed green, and this is the only check that validates the native-image metadata against the reshuffled codec jars, since each split artifact ships its own `META-INF/native-image` tree.
- The remaining Ballerina test packages beyond the two above.

## Behavioural changes to be aware of

- **The adaptive allocator is Netty 4.2's new default**, replacing pooled. This PR does not pin it. There are open upstream reports of direct-memory regressions ([#15850](https://github.com/netty/netty/issues/15850), [#17107](https://github.com/netty/netty/issues/17107), [#15767](https://github.com/netty/netty/issues/15767)), and this module's streaming path allocates direct buffers on the event loop and releases them on Ballerina worker threads, which is the pattern the adaptive allocator handles worst. The guide recommends rolling out with `-Dio.netty.allocator.type=pooled` and flipping separately once load-tested; that comparison has not been run yet.
- **HTTP/2 header value validation is on by default** as of 4.2.18, and chunked-must-be-last is validated regardless of HTTP version. `http2-tests` passes unchanged.
- **The pipeline call-stack was flattened**, which can change behaviour where the pipeline is modified concurrently. This module does that during H2C upgrade and the WebSocket handshake.
- `NioEventLoopGroup` is deprecated in favour of `MultiThreadIoEventLoopGroup` with an `IoHandlerFactory`. It still exists and is source-compatible, so migrating is left as a follow-up rather than mixed into this change.

## Security

Targets 4.2.18.Final, which is above the floor set by [GHSA-p85m-gvr3-788c](https://github.com/netty/netty/security/advisories/GHSA-p85m-gvr3-788c) (high): host name verification is disabled on the OpenSSL client path for `netty-handler` 4.2.0–4.2.16, and this module builds client contexts with `SslProvider.OPENSSL`.

## Tests

Adds six tests covering `verifyHostName` in both directions over HTTP/1.1 and HTTP/2, for both the trust-store and cert-file client configurations. `verifyHostName` was public API with no test exercising the disabled path. The tests reach the listener over `127.0.0.1` while the certificate is issued for `CN=localhost` with no SAN, which makes the setting observable without new fixtures, and assert that a failure is not a certification path error so the enabled cases can only pass on an actual host name mismatch.

## Summary

- Migrated the HTTP module to Netty 4.2.18.Final and tcnative 2.0.84.Final.
- Updated split Netty codec, DNS, resolver, compression, and OCSP dependencies.
- Removed unused marshalling and protobuf dependencies.
- Configured client TLS hostname verification explicitly.
- Added HTTP/1.1 and HTTP/2 hostname verification tests.
- Added native-image runtime initialization metadata.
- Improved embedded SOCKS server shutdown handling.
- Updated the crypto dependency and package changelog.


TharmiganK:
Duplicate of https://github.com/ballerina-platform/module-ballerina-http/pull/2709

TharmiganK:
## CodeRabbit Autofix Review Complete

Reviewed 1 CodeRabbit feedback item and did not apply code changes in this run.

The thread was anchored to `012c92755`, and the assertion it quotes has since been replaced in `ee9fad1b9` and `64d23af6f`. The enabled-verification assertions now require a positive host name diagnostic (`no subject alternative names`, `no name matching`, and other provider-specific wording) and fail on unrelated trust-path, protocol, or connection errors.

The HTTP/2 client is pinned to the OpenSSL provider, which reports a host name mismatch and a trust-path failure with an identical top-level message, so a message assertion cannot distinguish them there. That test instead proves the cause by contrast: the mismatched host must fail while the same configuration against the certificate's CN must succeed.

Thread resolved as already addressed.

PR #2717, opened 2026-09-21
Surface HTTP/2 content decoding failures instead of blocking the reader

### LINKED ISSUE #7566: JdkZlibDecoder ignores everything but first GZIP stream, non-compliant with RFC 1952
### Expected behavior
gzip content is fully consumed
### Actual behavior
gzip content is partially consumed
### Steps to reproduce
n/a
### Minimal yet complete reproducer code (or URL to code)
n/a
### Netty version
4.1.17
### JVM version (e.g. `java -version`)
n/a
### OS version (e.g. `uname -a`)
n/a

I'm using Spring WebClient to read a large (12 million records) gzipped response. I've set up the Netty pipeline as follows:
```
ctx.addHandlerLast(new ReadTimeoutHandler(readTimeoutMillis, TimeUnit.MILLISECONDS));
ctx.addHandlerLast(new Slf4JLoggingHandler());
if (forceDecompression) {
    io.netty.handler.codec.http.HttpHeaders httpHeaders = new ReadOnlyHttpHeaders(
            true,
            CONTENT_ENCODING, GZIP,
            CONTENT_TYPE, APPLICATION_JSON
    );
    HttpResponseHeadersHandler headersModifier = new HttpResponseHeadersHandler(httpHeaders);
    ctx.addHandlerFirst(headersModifier);
}
ctx.addHandlerLast(new HttpContentDecompressor());
```
The `HttpResponseHeadersHandler` overwrites headers in the response because the server doesn't send an `Content-Encoding` header. Details and implementation can be found [here](https://github.com/netty/netty/issues/7553#issuecomment-355249883).

Problem is, processing stops after about 130,000 records. No errors are shown. I've debugged and found out that the `JdkZlibDecoder` used for decompression stops reading and enters the finished block. I've tests that work fine for fewer number of records (100 or so).

However, using a different client (okhttp), if I don't set the `Content-Encoding` header and wrap the input stream with `GZIPInputStream`, everything works as expected.

So, question is, if `GZIPInputStream` doesn't have a problem, why does `JdkZlibDecoder`? And since it does, why won't it report what it thinks is the issue?

```
>> GET http://myserver.mycompany.com/.../businesses.20180104.json.gz
<< 200 OK
<< connection -> [keep-alive]
<< accept-ranges -> [bytes]
<< content-disposition -> [attachment; filename="businesses.20180104.json.gz"; filename*=UTF-8''businesses.20180104.json.gz] 
<< content-type -> [application/x-gzip]
<< content-length -> [3384998203]
<< date -> [Fri, 05 Jan 2018 00:43:32 GMT]
<< etag -> [0e49d5fa7ba9f68058bfbb4a98bef032c3a73871]
<< last-modified -> [Thu, 04 Jan 2018 23:54:26 GMT]
<< x-artifactory-id -> [9732f56568ea1e3d:59294f65:160b8066066:-8000]
<< x-checksum-md5 -> [451ca1b1414e7b511de874e61fd33eb2]
<< x-artifactory-filename -> [businesses.20180104.json.gz]
<< server -> [Artifactory/5.3.0]
<< x-checksum-sha1 -> [0e49d5fa7ba9f68058bfbb4a98bef032c3a73871]
```

For full disclosure, Initially posted a [question](https://stackoverflow.com/q/48094553/839733) on stackoverflow. Also related to https://github.com/square/okhttp/issues/3759.

### LINKED ISSUE #9191: [Bug] HTTP/2 client getByteStream() does not return when Content-Encoding: gzip body is malformed (HTTP/1.1 returns ClientError; timeout not observed)
### Area

module/http (Client, HTTP/2)

### Description

The HTTP/2 client does not return from `http:Response.getByteStream()` when a response declares `content-encoding: gzip` but the body is not valid gzip. No error is returned. With `timeout: 30`, no client timeout error was observed before external termination at 40 seconds.

The same response received over **HTTP/1.1 returns an `http:ClientError` from `getByteStream()` immediately**, so the failure path exists in the library but is not propagated on HTTP/2.

The same symptoms were observed on the JVM and in the GraalVM native executable (Ballerina Swan Lake 2201.13.6, `ballerina/http` 2.17.1), suggesting the issue is not specific to native-image compilation; the failing component is not yet established.

**RFC context:** per RFC 9113 §8.1, an HTTP/2 response is complete once a frame with the `END_STREAM` flag is received (including any CONTINUATION frames needed to finish the field block). Received `END_STREAM` closes the response direction; the stream state depends on whether the request direction has already ended (§5.1: from `open`, received `END_STREAM` is half-closed (remote); after a completed GET, the stream typically becomes `closed`). §6.1 requires any subsequent DATA on that stream to be treated as a `STREAM_CLOSED` error, so no further valid response DATA can arrive after `END_STREAM`. This report does not independently prove that the Ballerina client processed `END_STREAM` before the hang. RFC 9110 §8.4 identifies which decoding must be applied to obtain the indicated media type; it does not prescribe a Ballerina error type or timeout. Returning a decode failure to the caller is the requested library behavior, matching the HTTP/1.1 path.

### Steps to Reproduce

**1. Minimal TLS peer** (Node.js v26, `http2` with `allowHTTP1: true`, self-signed cert for `127.0.0.1`) that responds `200` with `content-encoding: gzip` and a non-gzip body:

```js
const http2 = require('node:http2');
const fs = require('node:fs');
const server = http2.createSecureServer({
  key: fs.readFileSync('key.pem'), cert: fs.readFileSync('cert.pem'), allowHTTP1: true
});
server.on('request', (request, response) => {
  const body = Buffer.from('not-a-gzip-stream'); // declared gzip, but plain bytes
  response.writeHead(200, {
    'content-type': 'application/json',
    'content-encoding': 'gzip',
    'content-length': body.length
  });
  response.end(body);
});
server.listen(9443, '127.0.0.1');
```

**2. Ballerina client** (Swan Lake 2201.13.6):

```ballerina
import ballerina/http;
import ballerina/io;

public function main(string endpoint, string certificate, string httpVersion = "2") returns error? {
    http:Client httpClient = check new (endpoint, {
        timeout: 30,
        httpVersion: httpVersion == "1" ? http:HTTP_1_1 : http:HTTP_2_0,
        retryConfig: {count: 0},
        secureSocket: {cert: certificate}
    });
    http:Response response = check httpClient->get("/badgzip");
    io:println("status=", response.statusCode);
    stream<byte[], io:Error?>|http:ClientError streamResult = response.getByteStream(8192);
    if streamResult is http:ClientError {
        io:println("STREAM_ERROR");
        return;
    }
    io:println("READING");
    while true {
        record {| byte[] value; |}|io:Error? item = streamResult.next();
        if item is io:Error { io:println("IO_ERROR"); return; }
        if item is () { break; }
    }
    io:println("DONE");
}
```

**3. Results** (each run bounded externally at 40s, JVM jar and native executable behave the same):

- `httpVersion = 2` → prints `status=200`, then **does not return from `getByteStream()` before external termination at 40s** (never reaches `READING`); killed externally, exit 124. Server logs confirm `ALPN=h2`.
- `httpVersion = 1` → prints `status=200`, `STREAM_ERROR`, exits 0. Server logs confirm `ALPN=http/1.1`.

Verified matrix against the same server (both JVM and native artifact):

| Response body | HTTP/1.1 | HTTP/2 |
|---|---|---|
| valid gzip (54 B) | decoded OK | decoded OK |
| valid gzip, streamed | decoded OK | decoded OK |
| valid gzip, 1 MiB | decoded OK | decoded OK |
| **malformed gzip (declared, plain bytes)** | `ClientError` immediately | **does not return (killed at 40s)** |
| truncated gzip | returns partial 31 B | returns partial 31 B |
| no response at all (stalled server) | UNVERIFIED in supplied logs | timeout error at ~30s |

The last row’s HTTP/2 stall timed out at ~30s (`Idle timeout triggered before initiating inbound response`); an HTTP/1.1 stall run was not in the supplied logs. That timeout does not establish that the same deadline covers body consumption or decode waits. The truncated-gzip row is a separate observation (partial decoded bytes returned as success); the sketch above implements only the malformed-body case.

### Affected Versions

Ballerina Swan Lake **2201.13.6** (latest stable at time of reporting), `ballerina/http` **2.17.1**. Both the JVM executable and the GraalVM native build (`bal build --graalvm`, GraalVM CE 21.0.2).

### Environment details

Linux x64 (Ubuntu-based container). Test peer: Node.js v26.8.2. Verified on JVM (GraalVM CE 21.0.2) and on the native-image executable built from the same package.

### Suggested fix

1. Surface the decoding failure as `http:ClientError` during `getByteStream()` construction, or as `io:Error` during stream iteration, matching the HTTP/1.1 failure path.
2. Ensure decoding failures terminate pending body operations, and clarify whether the configured timeout covers body consumption.
3. Secondary: a truncated gzip stream currently returns silently **partial** decoded content (31 of 34 bytes) as a successful read; an incomplete decode should arguably be an error as well.

### Impact

The tested mislabeled-gzip response caused the reported HTTP/2 hang: `getByteStream()` did not return and no timeout error was observed in a 40-second window. The same class of failure can occur if an endpoint or intermediary returns mislabeled gzip — for example a proxy injecting an error page under a copied `content-encoding: gzip` header. The captured HTTP/2 `/gzip` request contained no `accept-encoding` field (`:path`, `:scheme`, `:authority`, `:method`, `user-agent: ballerina` only); full headers were not captured for `/badgzip`. If no Accept-Encoding is sent, RFC 9110 §12.5.3 rule 1 ("If no Accept-Encoding header field is in the request, any content coding is considered acceptable") permits a content-coded response. HTTP 2.17.1 defaults `compression` to `COMPRESSION_AUTO`; that default does not override the `/gzip` capture.

TharmiganK:
## Purpose

Fixes ballerina-platform/ballerina-library#9191 (and its duplicate #9192).

When an HTTP/2 response declares `content-encoding: gzip` but the body is not valid gzip, the client reported nothing back to the caller and the response body read blocked. The same response over HTTP/1.1 returns an `http:ClientError` immediately.

Two corrections to the report, for the record:
- It is not an unbounded hang for the reader the reporter used. `BlockingEntityCollector` polls with `Constants.ENDPOINT_TIMEOUT` (300s), so the read eventually fails with `IllegalStateException("poll timeout expired")`. The reporter's window was 40s. Still a bug: 10x the configured `timeout: 30`, and a meaningless error.
- The client `timeout` never bounded body reads on *either* protocol; it only drives Netty's `IdleStateHandler`. That is a separate pre-existing defect, addressed here as well.

This is a long-standing design gap rather than a regression; it predates the 2021 repackaging of the transport.

While covering the other ways of reading the body from Ballerina (`getBinaryPayload`, `getTextPayload`, `getJsonPayload`, client data binding) I found two further places where a failed body did not reach the caller, on both protocols. They are fixed here too, see [Failures no longer lost or hidden](#failures-no-longer-lost-or-hidden).

## Root cause

The two protocols decode content in structurally different places, and only the HTTP/1.1 one had a route for the failure:

| | HTTP/1.1 | HTTP/2 |
|---|---|---|
| Decoder | `HttpContentDecompressor` in the pipeline, **before** `TargetHandler` | `DelegatingDecompressorFrameListener`, **inside** `Http2ConnectionHandler` |
| Failure | `DecoderException` via `exceptionCaught` | Netty `StreamException` |
| Who sees it | `TargetHandler.exceptionCaught` → `closeChannel` → `channelInactive` → `handleAbruptChannelClosure` → `StateUtil.handleIncompleteInboundMessage` | **Nobody.** `Http2ConnectionHandler.onStreamError` only writes a local RST_STREAM; it never calls `ctx.fireExceptionCaught`, and `Http2TargetHandler` has no `exceptionCaught` |

So on HTTP/2 `ClientFrameListener.onDataRead` never runs, no content or `LastHttpContent` reaches the response, `StreamCloseListener.onStreamClosed` dropped the locally initiated close silently, and `Http2ClientTimeoutHandler.onStreamClose` then cancelled the stream's idle timer, removing the last safety net.

The notification machinery already existed and was used for the *peer-sent* RST case (`ReceivingEntityBody.handleRstStream`); it was simply unreachable for locally generated stream errors.

For what it's worth, Vert.x uses the identical `DelegatingDecompressorFrameListener` model and closes this same hole by overriding `onStreamError` → `stream.onException(cause)`. Reactor Netty sidesteps it by putting `HttpContentDecompressor` in each HTTP/2 stream child pipeline instead of using a frame listener at all.

### The hang, fixed at two layers

- New `Http2ClientDecompressorFrameListener` wraps the Netty decompressor and reports a decode failure against the in-flight message with its real cause, so the error text matches HTTP/1.1. It also covers server push: a push response is tracked in the promised-message map rather than the in-flight map, so a malformed pushed body is reported to its own push response listener.
- `Http2ClientChannel.StreamCloseListener.onStreamClosed` now terminates any message still registered in flight when its stream closes. A completed response removes itself from that map first, so anything left is by definition incomplete. This backstops every locally aborted stream, not just gzip. It does not yet cover promised (push) streams; the decoder layer above does.

Both go through a new `SenderState.handleStreamClosedLocally`, implemented per state so the reason reflects where the exchange was.

**Avoiding double reporting:** `OutboundMsgHolder.claimStreamTermination()` is a compare-and-set latch letting exactly one of {received RST_STREAM, GOAWAY, idle timeout, connection close, local abort} notify the caller. An expired `Expect: 100-continue` wait deliberately does not claim it, since that timeout is not terminal and the request body is sent anyway. A push stream shares its parent request's holder, so it does not use the latch; leaving the promised map is what makes its notification single.

**Dead handler removed:** `configureHttp2Pipeline` added an `HttpContentDecompressor` *after* `Http2TargetHandler`, which consumes every frame type it knows and forwards nothing else, so it could never receive a message. Leaving it there made the HTTP/2 path look like it already handled decompression errors.

### Body-read timeout

Both response creation sites used the `HttpCarbonResponse` constructor that takes no wait time, pinning body reads to the 300s `ENDPOINT_TIMEOUT`. They now pass a wait derived from the socket idle timeout through `Util.resolveEntityWaitTime`, falling back to the endpoint default only when no timeout is configured.

The wait deliberately trails the configured timeout by `ENTITY_WAIT_GRACE_MILLIS` (5s). The idle timeout handlers end a stalled body at the configured timeout with the precise cause ("Idle timeout triggered while reading inbound response entity body"). A read bound equal to the timeout raced them, and when it expired first the reader got a generic "No entity was added to the queue before the timeout" instead; `ClientTimeoutWhileReadingBodyTestCase` failed on CI because of exactly that. With the grace the idle handlers always report first and the bound only takes effect when no report ever arrives.

### Failures no longer lost or hidden

Both of these are reachable on any inbound message whose body fails, not only on decode errors.

- **A failure that arrived before the payload accessor ran was dropped.** `getBinaryPayload`, `getTextPayload`, `getJsonPayload` and client data binding go through `getFullHttpCarbonMessage()`, which only creates its future when they are called. `notifyContentFailure()` reported to the future only if it already existed, and unlike success (latched in `lastHttpContentArrived`) nothing remembered the failure, so a fast failure left the accessor waiting forever on a listener that would never be called, with no timeout. The message now keeps the failure and replays it to a future created later, and `addListener` reports a recorded failure before checking for completion.
- **A failed body was treated as an empty one.** `HttpUtil.populateEntity` decides whether the entity has a body from `countMessageLengthTill(ONE_BYTE)` when there is no `content-length`. Netty's HTTP/2 decompressor removes `content-length` and, unlike the HTTP/1.1 one, does not switch the message to chunked, so the empty failing `LastHttpContent` counted as size 0. The entity was then set up without a body: `getBinaryPayload` returned an empty array and binding to `json` produced `null`, a silent success. A failed content is now counted as reaching the requested size, so the read that follows raises the failure.

> [!NOTE]
> Two behaviour changes are worth a release note, and both are in `changelog.md`:
> - The wait for an inbound response body now follows the configured client `timeout` (plus the grace) instead of a fixed 300s.
> - An HTTP/2 response that fails before any of its body arrives, for example on a reset or an idle timeout, now results in an error when its payload is read instead of an empty payload. That is the accurate outcome for an incomplete response, but it is observable.

## Known limitation, deliberately not fixed here

The tickets also note that a **truncated** gzip stream yields silent partial content. That affects HTTP/1.1 and HTTP/2 equally and is not fixable against Netty's public API without a worse trade:

| decoder setting | truncated | multi-member (RFC 1952) |
|---|---|---|
| `decompressConcatenated=true` (what `ZlibCodecFactory` always builds) | undetectable, `isClosed()` is permanently false | decodes both |
| `decompressConcatenated=false` | detectable | **silently drops everything after the first member** |

That flag was set to `true` by netty/netty@6ff48dc to fix [netty#7566](https://github.com/netty/netty/issues/7566), so flipping it here would re-introduce that bug for our users. Notably the JDK's own `GZIPInputStream` gets both right, which makes this a well-formed upstream Netty report with a reference implementation. Tracking separately.

Also not changed: the non-blocking payload accessor path (`ExternUtils.getResult`) waits on its future with no timeout of its own, so a notification that is lost by some other means would still leave it waiting, where the blocking read is bounded.

## Testing

Native (transport) tests:
- `ClientRespDecodingFailureTestCase` drives a Netty peer serving a mislabeled gzip body over both protocols, a malformed **push** response over HTTP/2, and a valid-gzip control, and asserts the read fails with a `DecoderException` rather than blocking. With the fix reverted the HTTP/2 cases fail (the read never returns within 15s) while the HTTP/1.1 one still passes, which is the asymmetry in the ticket.
- `InboundResponseWaitTimeTestCase` covers the bound on a body read and that it trails the configured timeout.
- `FullHttpMessageContentFailureTestCase` covers replaying a failure to a late listener, a recorded failure winning over completion in either order, and a failed body not looking empty. The replay, ordering and empty-body cases each fail without their fix.

Ballerina tests (`http2-tests`, group `gzipDecodeFailure`), 23 passing:
- A Ballerina service cannot emit a mislabeled gzip body, because `HttpUtil.setCompressionHeaders` overwrites `content-encoding` on every outbound response, so these run against a Java server added to `test-utils` (Netty, HTTP/1.1 and HTTP/2 prior knowledge, with a push path).
- The malformed body is read through `getByteStream`, `getBinaryPayload`, `getTextPayload`, `getJsonPayload` and client data binding to `byte[]`, `string` and `json`, over HTTP/2 with an HTTP/1.1 control, and each must fail within 20s. The same accessors decode a valid body, the client keeps working on the same connection after a failed read, and a malformed pushed response fails without affecting the response that carried the promise.
- These are what exposed the two "no longer lost or hidden" defects: before those fixes `getBinaryPayload` and binding to `byte[]`/`json` returned an empty payload or hung.

Verification run locally: the whole `http2-tests` package passes (185), and 86 transport tests across the timeout, idle, back-pressure, passthrough, compression, streaming and content-aware classes pass. The full transport suite was green (532/532) at the first revision of this PR; after the follow-ups I ran the related classes rather than the whole suite, so CI covers the remainder. `BackPressureAwareIdleStateHandlerTest` is timing sensitive (300ms idle timeout with a 60ms margin) and failed intermittently when run alone on a loaded machine; it does not touch this change.

## Examples

N/A

## Checklist
- [x] Linked to an issue
- [x] Updated the changelog
- [x] Added tests
- [ ] Updated the spec

## Summary

- Fixed HTTP/2 response-body reads that block on malformed content encoding.
- Propagated decoding failures through streams, payload accessors, data binding, and pushed responses.
- Completed pending operations when HTTP/2 streams close locally.
- Prevented duplicate stream-termination notifications.
- Applied the configured client timeout, with a grace period, to response-body reads.
- Ensured failed bodies are not treated as empty and replayed failures to later payload accessors.
- Refactored HTTP/2 decompression handling.
- Added coverage for malformed and valid gzip responses, push responses, body accessors, data binding, connection reuse, and timeout behavior.


native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/sender/http2/Http2ClientDecompressorFrameListener.java
```diff
@@ -0,0 +1,137 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (http://www.wso2.org) All Rights Reserved.
```
daneshk:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (http://www.wso2.org).
```

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/sender/http2/Http2ClientDecompressorFrameListener.java
```diff
@@ -0,0 +1,137 @@
...
+        super(connection, clientFrameListener);
+        this.clientFrameListener = clientFrameListener;
+    }
+
+    @Override
+    public void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers headers, int padding,
+                              boolean endStream) throws Http2Exception {
+        try {
+            super.onHeadersRead(ctx, streamId, headers, padding, endStream);
+        } catch (Exception cause) {
```
daneshk: Is there a possible to get other error than decoding failures here. g (e.g. a malformed HPACK header block)

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/sender/states/http2/EntityBodyReceived.java
```diff
@@ -128,4 +128,9 @@ public void handleServerGoAway(OutboundMsgHolder outboundMsgHolder) {
     public void handleRstStream(OutboundMsgHolder outboundMsgHolder) {
         LOG.warn("handleRstStream is not a dependant action of this state");
     }
+
+    @Override
+    public void handleStreamClosedLocally(OutboundMsgHolder outboundMsgHolder) {
+        LOG.debug("handleStreamClosedLocally is not a dependant action of this state");
```
daneshk: Shall we reword this message than using the method name `handleStreamClosedLocally`

native/src/main/java/io/ballerina/stdlib/http/transport/message/BlockingEntityCollector.java
daneshk, line 118: Don't we need to handle the failure in decoderResult here as well?

test-utils/src/main/java/io/ballerina/stdlib/http/testutils/GzipHttp2ConnectionHandler.java
```diff
@@ -0,0 +1,136 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com) All Rights Reserved.
```
daneshk:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
```

daneshk: please check other places

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/sender/http2/Http2ClientDecompressorFrameListener.java
TharmiganK: Applied literally, and also normalized to `WSO2 LLC. (http://www.wso2.com).` on the other 11 new files I'd copy-pasted the wrong header into (also flagged below) — that's the form the newest files elsewhere in the repo already use. In 1b806b3f7.

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/sender/http2/Http2ClientDecompressorFrameListener.java
TharmiganK: Good catch, no — `onHeadersRead` never actually decompresses anything here. `DelegatingDecompressorFrameListener` only decompresses in `onDataRead`, through `Http2Decompressor.decompress`; its `onHeadersRead` just sets up the per-stream decompressor and forwards to the delegate listener (our `ClientFrameListener`). So my `onHeadersRead` override was actually catching that delegate's own header-handling exceptions and mislabeling them "Content decoding failed", not HPACK or decoding errors (HPACK itself is already decoded into `Http2Headers` before this listener ever sees it). Removed both `onHeadersRead` overrides and kept only `onDataRead`, with the javadoc explaining why. In 1b806b3f7.

native/src/main/java/io/ballerina/stdlib/http/transport/contractimpl/sender/states/http2/EntityBodyReceived.java
TharmiganK: Reworded to "Stream closed locally after the entity body was already received; no action needed" in 1b806b3f7.

native/src/main/java/io/ballerina/stdlib/http/transport/message/BlockingEntityCollector.java
TharmiganK: Yes, missed that one - `countMessageLengthTill` handled it but `getFullMessageLength` didn't, so a failed decode with no further `LastHttpContent` would sit waiting out the full `soTimeOut` on an empty queue before giving up silently. Mirrored the same check there in 1b806b3f7.

test-utils/src/main/java/io/ballerina/stdlib/http/testutils/GzipHttp2ConnectionHandler.java
TharmiganK: Applied, and checked the other new files - see the reply above.

TharmiganK: Checked every file this PR added: 9 more had the same `wso2.org` / "All Rights Reserved." mistake (the 3 in test-utils already matched this file). All 12 are now `Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).`, in 1b806b3f7.

daneshk (approved):

